PR #12, opened 2020-06-05
Add Githubs Change Password URL

dikti:


rmondello:
Github appears to have adopted the Well-Known URL for Changing Passwords, so I’m not sure if it’s appropriate to incorporate a quirk. There’s no need! @dikti, does that make sense?

rmondello:
For reference: https://github.com/.well-known/change-password

dikti:
You’re right, my bad.

bradcush:
@rmondello Are you aligned with creating a list of sites that support the "Well-Known URL for Changing Passwords"?

rmondello:
Good question, @bradcush. I don’t think that falls within the spirit of the project, which is to gather data to help password managers help users. Listing out website that do support the well-known change password URL doesn’t really do that. Does that make sense? What do you think?

bradcush:
@rmondello Makes sense, thanks!

PR #18, opened 2020-06-05
add williams-sonoma inc sites with shared backends

bradcush:
Add Williams-Sonoma Inc. partner sites with shared backends using the same login credentials

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -289,6 +289,12 @@
         "wikimania.wikimedia.org"
     ],
     [
+        "williams-sonoma.com",
+        "markandgraham.com",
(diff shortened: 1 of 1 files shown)


geoffcline:
I double checked these retail sites shared a backend.

rmondello:
I’m working through some of the older PRs. Some comments:

1. This needs to be updated to merge cleanly.
2. And we’re sure that each of these domains serves a login page, rather than redirecting to one login page on one domain?

bradcush:
@rmondello Apologies for the delayed response... I've updated the PR so it can be merged cleanly. When I created it I also verified that each domain I've listed serves it's own login page for it's own domain, and that it's not shared.

rmondello:
Thanks @bradcush!

PR #32, opened 2020-06-05
Add facebook.com change password URL

gobinathm:
- Add Facebook.com password change URL


hober:
I tested today, and Facebook supports the well-known URL for changing passwords. So I don't think we need to add it to this project, which lists change password URLs for sites that *don't* support the well-known URL for changing passwords.

PR #34, opened 2020-06-05
Remove twitter.com change password URL

tregagnon:
Twitter.com supports `.well-known/change-password`. You can try it: https://twitter.com/.well-known/change-password

As explained in the Project's README, those websites should not be included in this file.

Also remove the duplicated line for fnac.com.


rik:
> I think we might want to clarify in the README that alphabetical sorting should be used in these files, where applicable. That’ll help us with some of the merge conflicts.

Maybe a bot could alter PRs to sort them and make it easier to contribute?

rmondello:
Good idea, @rik. I’ll see if we can make that happen.

PR #38, opened 2020-06-05
Added Autodesk's change password URL

valentinvarbanov:
Checked and autodesk.com/.well-known/change-password does not work


rmondello:
Thanks @valentinvarbanov!

PR #40, opened 2020-06-05
Remove Amazon.co.jp from websites-with-shared-credential-backends.json

Hamuko:
Amazon.co.jp is on a different platform than Amazon.com and the rest of
Amazon sites and thus requires a different login.

https://www.amazon.co.jp/b?node=5835715051


rmondello:
Wow! Thanks!

PR #43, opened 2020-06-05
Add myaccess.dmdc.osd.mil to password rules

ewc7092:
The myaccess.dmdc.osd.mil site can be used by veterans and their families to access the eBenefits site.  The password requirements are complex and have a much shorter password rotation period, making this a perfect candidate for these efforts.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -141,6 +141,9 @@
     "metlife.com": {
         "password-rules": "minlength: 6; maxlength: 20;"
     },
+    "myaccess.dmdc.osd.mil": {
+        "password-rules": "minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@_#!&$`%*+()./,;~:{}|?>=<^[]'-];"
+    },
     "myhealthrecord.com": {
         "password-rules": "minlength: 8; maxlength: 20; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789_.!$*=];"
     },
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -141,6 +141,9 @@
     "metlife.com": {
         "password-rules": "minlength: 6; maxlength: 20;"
     },
+    "myaccess.dmdc.osd.mil": {
+        "password-rules": "minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@_#!&$`%*+()./,;~:{}|?>=<^[]'-];"
```
rmondello, line 145: This isn’t a valid rule, because a `]` has to be the last character in a character class. Which gives us:

```
minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@_#!&$`%*+()./,;~:{}|?>=<^'-[]];
```

But the fact the lower, upper, and digit are required means that they don’t need to be stated in allowed. Which gives us:

```
minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; allowed: [@_#!&$`%*+()./,;~:{}|?>=<^'-[]];
```

And just to be sure, all of those symbols are allowed, and no symbols is required, right?

rmondello:
Cool! If you could update the PR something that’s been somewhat simplified to not list out the alphabet twice and the ten digits, I’d be happy to merge!

ewc7092:
Ok, should be good to go now!

rmondello:
Thanks @ewc7092!

MaivKhanh:
Khánh Mai liked your email

PR #44, opened 2020-06-05
add stackexchange sites with shared backends

bradcush:
From what I could find these are all the main websites that are part of the stackexchange network that have their own second level domain.

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -176,6 +176,13 @@
         "skiheavenly.com"
     ],
     [
+        "stackoverflow.com",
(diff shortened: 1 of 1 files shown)


igor-makarov:
Could you please insert your entry in top-level alphabetical order? 

Preferably, by using stackoverflow.com as the first domain as it's the most popular.

bradcush:
@igor-makarov I've moved the list under the entry for `snow.com` and placed `stackoverflow.com` as the first entry list.

rmondello:
@bradcush @igor-makarov Thank you both for all of the contributions and reviews so far!

rmondello:
@Booligoosh And for reviewing this!

PR #45, opened 2020-06-05
Add key.harvard.edu to password rules

ewc7092:
This commit adds key.harvard.edu to the password ruleset.  Verified with the password change utility.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -141,6 +141,9 @@
     "metlife.com": {
         "password-rules": "minlength: 6; maxlength: 20;"
     },
+    "myaccess.dmdc.osd.mil": {
+        "password-rules": "minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@_#!&$`%*+()./,;~:{}|?>=<^[]'-];"
+    },
     "myhealthrecord.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -102,6 +102,9 @@
     "girlscouts.org": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$#!];"
     },
+    "key.harvard.edu": {
+        "password-rules": "minlength: 10; maxlength: 100; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@_#!&$`%*+()./,;~:{}|?>=<^[]'-];"
```
rmondello, line 106: In a character class (`[]`), the `]` has to be the last character in the set. So this rule doesn’t look like it validates on the Password Rules Validation Tool.

I think that what you were going for is equivalent to:

```
minlength: 10; maxlength: 100; required: lower; required: upper; required: digit; allowed: [@_#!&$`%*+()./,;~:{}|?>=<^['-]];
```

rmondello: How did you figure out what the maxlength and allowed special characters are? Experimentation, or something stated on the website?

ewc7092: Thanks (again!) for providing the corrected rule, I've uploaded the changes and I'll keep that in mind for future commits. 

The string rules at key.harvard for min, max length are outlined on the password reset page (which I was shocked to see a 100 char max, good on them!). The allowed special character list is not defined so I had to experiment.

rmondello: Sounds like a lot of work! But I bet people will appreciate it!

PR #46, opened 2020-06-05
Add HN password change URL

c0dr:


Hamuko:
Isn't this the same as #46?

rmondello:
Thanks!

PR #47, opened 2020-06-05
Add nintendo.com password rules

tregagnon:


rmondello:
This is an interesting case.

```
minlength: 8; required: lower; required: upper; required: digit;
```

This is a very reasonable set of requirements. I suspect they’re trivially satisfied by most password generation schemes. I wonder if a set like this is worth incorporating. I’d love to discuss this.

tregagnon:
Fair point! I hope password managers have similar (and even better) default generation settings. I agree with you. There is little value to add those very common requirements to the file.

But users can also change the password generation settings, and define them with lower values than this common set. How can we help password managers be smarter about it, and recognize than the website requires better than that? Something to think about, I don't have good solutions for now. 

You can discard the PR if you want. :)

rmondello:
I’m still thinking about this. We should set a bar for “this website’s policies are overall reasonable, and therefore, don’t need to be documented in this way”, or we could rely on adopters of this data set to filter out rules that are trivially satisfied by their generation schemes. We could even do both!

I’ll take a little more time to think about this. Would be interested in hearing other folks’ opinions, too.

rmondello:
Actually, the rule is incorrect. For instance, nintendo.com allows a `-` in the password, but this rule says this can’t be the case. Something may still be worth including here, or not, but not as-is.

rmondello:
They seem to allow every symbol I can experiment with other than `'` and `"`.

rmondello:
Closing this due to inactivity. Feel free to re-open or re-submit! We’d love your contribution.

PR #50, opened 2020-06-05
Add key.harvard.edu change password url

ewc7092:
This change adds the key.harvard.edu password change url for easier password management.


rmondello:
Thanks @ewc7092!

PR #52, opened 2020-06-05
Added StackOverflow and StackExchange

cbjeukendrup:


igor-makarov:
#44 looks more complete, I think. Perhaps close this PR in favor of it?

PR #53, opened 2020-06-05
Adopt mandatory order #15

### LINKED ISSUE #15: Adopt mandatory order
While the files currently are relatively short, they will probably fill up with rules in no time. To maintain some sort of clarity, I propose to add a mandatory order, like alphabetical, to the contribution rules.

rmondello:
This resolves #15.


quirks/change-password-URLs.json
```diff
@@ -1,9 +1,9 @@
...
-    "tripit.com": "https://tripit.com/account/edit/section/change_password",
-    "twitch.tv": "https://www.twitch.tv/settings/security",
     "ea.com": "https://myaccount.ea.com/cp-ui/security/index",
     "fnac.com": "https://secure.fnac.com/account/update-password",
+    "impots.gouv.fr": "https://cfspart.impots.gouv.fr/monprofil-webapp/GererMonProfil",
     "microsoft.com": "https://account.live.com/password/Change",
     "nintendo.com": "https://accounts.nintendo.com/password/edit",
-    "impots.gouv.fr": "https://cfspart.impots.gouv.fr/monprofil-webapp/GererMonProfil"
+    "tripit.com": "https://tripit.com/account/edit/section/change_password",
+    "twitch.tv": "https://www.twitch.tv/settings/security"
```
darinadler, line 8: Does JSON allow a trailing comma on the last element? If it does, I would suggest including it for easier merging.

rmondello: JSON disallows trailing commas, but there may be workarounds or a format change that could help us. I plan on investigating this after working through the initial set of pull requests!

PR #54, opened 2020-06-05
Add all eBay sites

igor-makarov:
All the local eBay sites use different TLDs. You can check the SSL certificates to verify this.


rmondello:
Oh, this is great. Thanks @igor-makarov!

PR #56, opened 2020-06-05
Add myaccount.ea.com and origin.com to shared credential backends

DanTheMan827:


quirks/websites-with-shared-credential-backends.json
```diff
@@ -198,5 +198,9 @@
     [
         "lookmark.io",
         "lookmark.link"
+    ],
+    [
+        "origin.com",
+        "myaccount.ea.com"
```
bradcush, line 204: When trying to access `myaccount.ea.com` it doesn't seem to resolve. Also signing in to `origin.com` prompts a new window that references `signin.ea.com`. (`ea.com` also redirects you to sign in at `signin.ea.com`)

igor-makarov:
Perhaps consider changing it to just `ea.com` which would cover all subdomains?

erynofwales:
Yeah, I see the same thing when I try to log into origin.com. It pops a new window with an ea.com URL. Since we're not logging into origin.com, I don't think it makes sense to incorporate this.

PR #57, opened 2020-06-05
Add change password url for myaccount.ea.com

DanTheMan827:


DanTheMan827:
This is also the URL that origin.com redirects to #56

PR #58, opened 2020-06-05
add turbotax and gap sites to shared credential backends

hmpatel:


quirks/websites-with-shared-credential-backends.json
```diff
@@ -198,5 +199,13 @@
...
         "lookmark.link"
+    ],
+    [
+        "gap.com",
+        "bananarepublic.com",
+        "athleta.com",
+        "hillcity.com",
+        "oldnavy.com"
     ]
+
```
rmondello, line 210: Let’s remove this errant newline, please!

hmpatel: I logged in to gap.com using my credential and then verified through their listed partner sites and it all worked on SSO methodology using a single account.

hmpatel: More details from GAP sign in page -  https://secure-www.gap.com/profile/sign_in.do

image

geoffcline:
I can confirm the 5 brands (in the image above) have different domains and the same backend

rmondello:
But they all appear to sign in at something.gap.com, which means that it’s not necessary for password managers to expose credentials saved for any of those websites to each other. What do y’all think?

Maybe we should clarify the purpose of this file a little bit more clearly, or add a clarifying remark to the PR template?

geoffcline:
Ah that is an interesting point. 

I assumed that {gap; old navy; banana republic} handled it the same as {mens warehouse; jos a bank}, and {wiliams sonoma; west elm; pottery barn}. 

#18 (wiliams sonomo) is necessary -- they have different URLs
https://www.williams-sonoma.com/account/login.html
https://www.westelm.com/account/login.html

this isn't necessary for gap, they have the same base domain -- see, for ex,
https://secure-www.gap.com/profile/sign_in.do
https://secure-bananarepublic.gap.com/profile/sign_in.do

I also checked for the gap the experience of signing in on iOS with 1Password. It looks like all their iOS apps have "secure-www.gap.com" as an associated domain. (ref, https://developer.apple.com/documentation/safariservices/supporting_associated_domains_in_your_app) 

I think this is worth mentioning in the pull request template -- perhaps I will pull request the pull request? :)

hmpatel:
This makes a lot of sense now. I think I got carried away looking at some of the existing websites in this file. The revised PR template is perfect for new contributors.  

Based on this, from a cursory glance I see a few sites which requires cleanup - e.g. {comcast.net, xfinity.com}  both redirect to my.xfinity.com, similarly for {dropbox.com, getdropbox.com} and  {dish.com, mydish.com, dishnetwork.com} etc.. to the same parent domain

rmondello:
> Based on this, from a cursory glance I see a few sites which requires cleanup - e.g. {comcast.net, xfinity.com} both redirect to my.xfinity.com, similarly for {dropbox.com, getdropbox.com} and {dish.com, mydish.com, dishnetwork.com} etc.. to the same parent domain

Ah, then there is one criteria that is missing to consider! In these cases, the websites used to allow signing in on the other domain, and maintaining them on this list (or a list that is clarified to serve this purpose) can still be useful for correlating credentials within a password manager, or not warning about password reuse across them. (Say I created my Dropbox account at getdropbox.com and used it for a while before they switched over to dropbox.com.)

hmpatel:
Sorry had off on this for sometime...So kind of still trying to make sense here - so for instance dishnetwork.com which is basically an alias resolves to dish.com. 

This is unlike the scenario where 2 different domains allowing to sign in using same credentials
 https://www.williams-sonoma.com/account/login.html
https://www.westelm.com/account/login.html 

There could be a lot of aliases and CNAMEs for a domain and mapping all of them would proliferate this file beyond control. I guess my point is do we want aliases for the same domain mapped in this file ? Isn't the intent to map different domains sharing the same credentials backend to be kept in this file?

rmondello:
@hmpatel If the websites each serve a login page that can be logged into with the same credentials (as opposed to redirecting through one domain), then they belong on this list!

rmondello:
@hmpatel Happy to help you get this merged if you can fix up the merge conflict.

hmpatel:
@rmondello will close this PR as turbotax.com also serves off the same base domain i.e. x.intuit.com

rmondello:
@hmpatel Although it didn’t pan out, I appreciate you working through this pull request.

PR #60, opened 2020-06-05
Add aeroplan password rules

justin0022:
Screen Shot 2020-06-05 at 3 48 26 PM

Testing notes: I tried with 32 character password, which worked. I tried with 75 character password, which *did not work*. 

The password field copy/paste is disabled, which makes it very annoying to test (I used DevTools to fill my passwords in).

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -8,6 +8,7 @@
     "geocaching.com": "https://www.geocaching.com/account/settings/changepassword",
     "impots.gouv.fr": "https://cfspart.impots.gouv.fr/monprofil-webapp/GererMonProfil",
     "microsoft.com": "https://account.live.com/password/Change",
+    "myaccount.ea.com": "https://myaccount.ea.com/cp-ui/security/index",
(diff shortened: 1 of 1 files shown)


rmondello:
This website has a pretty reasonable set of rules. I don’t think a lot of password generators would fail to satisfy them. I wonder if this is worth incorporating. (The change looks totally correct, by the way!)

If anyone has opinions on this, I’d like to discuss it.

rmondello:
Oh, but the fact that they don’t allow any symbols (which your rule claims), actually does make it notable. Is it true that they don’t allow any symbols?

justin0022:
Does my rule claim that it doesn't allow any symbols? 

Hmm ya good point about it being reasonable. I think they changed it recently, before it only allowed [6-10 characters](https://password-shaming.tumblr.com/post/132422907286/aeroplan-aeroplan-is-basically-the-canadian).

Could there be some way to indicate websites that disable copy/paste into the password fields? There should be a special place in website hell for those sites...

justin0022:
Apologies for closing - it wasn't my intent. I force-pushed to update my branch to make a couple unrelated PRs and it seems to have closed this...

PR #62, opened 2020-06-05
add gov.br

rafaelvannucci:
Add Brazilian government unified services portal

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -103,8 +103,7 @@
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$#!];"
     },
     "gov.br": {
-        "password-rules": "minlength: 8; maxlength: 70; required: [!"#$%&()*+,-./:;<=>?@[\]^{|}~.];"
-        "exact-domain-match-only": true
+        "password-rules": "minlength: 8; maxlength: 70;"
     },
     "hawaiianairlines.com": {
         "password-rules": "maxlength: 16;"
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -102,6 +102,10 @@
     "girlscouts.org": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$#!];"
     },
+    "gov.br": {
+        "password-rules": "minlength: 8; maxlength: 70; required: [!"#$%&()*+,-./:;<=>?@[\]^{|}~.];"
```
rmondello, line 106: I don’t think you can include a quote here like this. This needs to be escaped, somehow.

quirks/password-rules.json
```diff
@@ -102,6 +102,10 @@
...
+        "exact-domain-match-only": true
```
rmondello, line 107: So you’re trying to say that no subdomains of gov.br should have these rules?

rafaelvannucci: edited to match actua rules.

quirks/password-rules.json
rafaelvannucci: edited to match actua rules.

rmondello:
> `minlength: 8; maxlength: 70;`

If that’s the websites requirements, it’s one of the most flexible websites I’ve ever seen! I’m not sure what value there is in tracking a maxlength of 70 for this website, but I’d be happy to hear arguments about this!

rafaelvannucci:
> > `minlength: 8; maxlength: 70;`
> 
> If that’s the websites requirements, it’s one of the most flexible websites I’ve ever seen! I’m not sure what value there is in tracking a maxlength of 70 for this website, but I’d be happy to hear arguments about this!

You're right @rmondello :/ It's very flexible. I'll go ahead and close this PR, as I don't have any argument about that.

PR #65, opened 2020-06-06
Include password rule for `amazon.com`

ameya-pandilwar:


rmondello:
Given that no password manager would recommend someone use a password with fewer than six characters, I wonder if this is actually worth tracking in the project.

ameya-pandilwar:
@rmondello  -- There is one more site that is being tracked currently --
https://github.com/apple/password-manager-resources/blob/0e4480cf01ed8692f6950ab1d753d74b0c846b78/quirks/password-rules.json#L27

Since, `amazon.com` is a widely used site, it doesn't hurt to include this rule. This is just a base guideline and not a strict enforcement. The user can ultimately create a password as desired.

PR #66, opened 2020-06-06
Add reddit.com

DanTheMan827:


quirks/change-password-URLs.json
```diff
@@ -11,6 +11,7 @@
     "myaccount.ea.com": "https://myaccount.ea.com/cp-ui/security/index",
     "news.ycombinator.com": "https://news.ycombinator.com/changepw",
     "nintendo.com": "https://accounts.nintendo.com/password/edit",
+    "reddit.com": "https://www.reddit.com/settings",
```
rmondello, line 14: This is a 404 for me.

From what I can tell, the URL we want is:
https://www.reddit.com/prefs/update/

quirks/password-rules.json
```diff
@@ -162,6 +162,9 @@
     "propelfuels.com": {
         "password-rules": "minlength: 6; maxlength: 16;"
     },
+    "reddit.com": {
+        "password-rules": "minlength: 6;"
```
rmondello, line 166: Given that no password manager I’m aware of would attempt to, by default, generate a password with fewer than 6 characters, I’m not sure this is worth tracking in the project. Like, I don’t think it’ll have a meaningful effect.

PR #70, opened 2020-06-06
Add gog.com

DanTheMan827:


quirks/password-rules.json
```diff
@@ -102,6 +102,9 @@
     "girlscouts.org": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$#!];"
     },
+    "gog.com": {
+        "password-rules": "minlength: 2;"
```
rmondello, line 106: I don’t think this is worth tracking in the project. A website requiring at least two characters in a password isn’t notable. Can you remove this?

quirks/change-password-URLs.json
```diff
@@ -6,6 +6,7 @@
     "ea.com": "https://myaccount.ea.com/cp-ui/security/index",
     "fnac.com": "https://secure.fnac.com/account/update-password",
     "geocaching.com": "https://www.geocaching.com/account/settings/changepassword",
+    "gog.com": "https://www.gog.com/account/settings/security",
```
rmondello, line 9: Looks good!

PR #71, opened 2020-06-06
Added sunlife

justin0022:
Screen Shot 2020-06-05 at 8 00 20 PM
8-10 characters. **8-10!**

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -189,6 +189,9 @@
     "ssa.gov": {
         "password-rules": "required: lower; required: upper; required: digit; required: [~!@#$%^&*];"
     },
+    "sunlife.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -189,6 +189,9 @@
     "ssa.gov": {
         "password-rules": "required: lower; required: upper; required: digit; required: [~!@#$%^&*];"
     },
+    "sunlife.com": {
+        "password-rules": "minlength: 8; maxlength: 10; required: digit; allowed: upper, lower;"
```
rmondello, line 193: The website’s rules you quote say the site must have at least one letter, but this only says that letters are allowed. I think this should be:

```
minlength: 8; maxlength: 10; required: digit; required: upper, lower;
```

justin0022: Thanks, I've fixed this now!

rmondello:
Thanks!

PR #73, opened 2020-06-06
Remove duplicate required: digit; for vanguard

justin0022:
I noticed there were two `required: digit;` and thought it was a typo.


rmondello:
Nice eye, @justin0022. I am, however, about to blow your mind.

I went digging; I added this rule before the project was released. I found this image:
image

Not a typo!!

It feels like, maybe, we should specify a “comment” key for entries, which can be used to document oddities like this.

justin0022:
Ah thanks for clarifying! A "comment" key sounds like an idea worth exploring.

PR #78, opened 2020-06-06
Add rules for bankofamerica.com

cglong:
Screenshot of BofA's website listing passcode rules:

image

Technically special characters aren't subject to the "more than 3 times in a row" rule, but I don't think the syntax supports this case 🙂


rmondello:
Thanks @cglong!

PR #85, opened 2020-06-06
Add acrons.com Password change URL

gobinathm:
- [X] There is no Well-Known URL for Changing Passwords (https://example.com/.well-known/change-password)

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -1,5 +1,6 @@
 {
     "500px.com": "https://web.500px.com/settings/account/security",
+    "acorns.com" "https://app.acorns.com/settings/change-password,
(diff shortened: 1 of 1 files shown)


quirks/change-password-URLs.json
```diff
@@ -1,5 +1,6 @@
 {
     "500px.com": "https://web.500px.com/settings/account/security",
+    "acorns.com" "https://app.acorns.com/settings/change-password,
```
rmondello, line 3: You need a closing quote on this line. :)

gobinathm: Thanks, @rmondello.

I did not realize i made such a silly mistake 👎 
Correct it

rmondello: No worries! Mistakes happen!

PR #86, opened 2020-06-06
Include rule for github.com

ameya-pandilwar:


igor-makarov:
These look like pretty common-sense requirements that any good password manager adheres to. 

The intention of this repo is to document sites that do not conform to the industry best practices.

djbe:
@igor-makarov the github.com one should be added though, because of the `required: digit` part.

ameya-pandilwar:
Hi @igor-markov -- I agree with what @djbe is saying and we should include `github.com`

quirks/password-rules.json
```diff
@@ -120,6 +120,9 @@
     "girlscouts.org": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$#!];"
     },
+    "github.com": {
+        "password-rules": "minlength: 8; required: lower; required: digit;"
```
rmondello, line 124: Github allows a much larger character set than just lower and digit, so this rule is incomplete. I suspect that Github isn’t really disallowing any specific characters, and their rules seem pretty reasonable, so I’m not sure they’re worth including in this project. (I am totally open to discussion about this point, however!)

djbe: GitHub password rules (https://help.github.com/en/github/authenticating-to-github/creating-a-strong-password):
- Eight characters long, if it includes a number and a lowercase letter, or
- at least 15 characters long with any combination of characters

This is maybe a more general discussion for this project: do we want to document every password quirk for each website? Or can we agree to a set of rules that we **assume** that each password manager will at least adhere to? I assume that each password manager will generate something longer than 8 characters. Not sure if I can assume the same for 16 chars (or longer), that's something that may be dependent on the user's settings.

rmondello: I think we do want to do this! I think we want to define such a policy, document it in the README, and reference it in the PR checklist. I opened https://github.com/apple/password-manager-resources/issues/134 to track this. (I’d be happy for anyone to take a crack at it.)

ameya-pandilwar: @rmondello -- Updated to include the allowed character set.

quirks/password-rules.json
```diff
@@ -120,6 +120,9 @@
     "girlscouts.org": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$#!];"
     },
+    "github.com": {
+        "password-rules": "minlength: 8; required: lower; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`!@#$%^&*()+~{}'\";:<>?];"
```
rmondello, line 124: This is equivalent to the easier to read: `minlength: 8; required: lower; required: digit; allowed: lower, upper, digit, [`!@#$%^&*()+~{}'\";:<>?];`

(I should fix the girlscouts.org rule, which is currently setting a bad example.)

ameya-pandilwar: @rmondello -- Thanks for the suggestion. I fixed it (along with the girlscouts.org rule as well)

PR #93, opened 2020-06-06
Add NYTimes Password Change URL

gobinathm:
### Checklist
- [X] There are no [open pull requests](https://github.com/apple/password-manager-resources/pull) for the same update.


rmondello:
Thank you!

PR #96, opened 2020-06-06
Added some German banks

DerJuulsn:
Add password rules for:

- HypoVereinsbank
- Deutsche Bank
- Consors Finanz
- Consors Bank
- Commerzbank
- Santanter


quirks/change-password-URLs.json
```diff
@@ -2,6 +2,7 @@
     "500px.com": "https://web.500px.com/settings/account/security",
     "amctheatres.com": "https://www.amctheatres.com/amcstubs/account",
     "autodesk.com": "https://accounts.autodesk.com/Profile/Security",
+    "deutsche-bank.de": "https://www.deutsche-bank.de/pfb/content/marktinformationen/passwort_vergessen.html",
```
rmondello, line 5: Is this a forgot password link or a change password link?

quirks/change-password-URLs.json
```diff
@@ -13,6 +14,7 @@
     "myaccount.ea.com": "https://myaccount.ea.com/cp-ui/security/index",
     "news.ycombinator.com": "https://news.ycombinator.com/changepw",
     "nintendo.com": "https://accounts.nintendo.com/password/edit",
+    "santander.de" : "https://meine.santander.de/login/#/passwort-zuruecksetzen",
```
rmondello, line 17: Same question here.

quirks/password-rules.json
```diff
@@ -53,6 +53,15 @@
     "comcastpaymentcenter.com": {
         "password-rules": "minlength: 8; maxlength: 20; required: lower, upper; required: digit; max-consecutive: 2;"
     },
+    "commerzbank.de": {
+        "password-rules": "required: upper,lower; required: digit; minlength: 5; maxlength: 8;"
+    },
+    "consorsbank.de": {
+        "password-rules": "required: upper,lower, digit; minlength: 5; maxlength: 5;"
```
rmondello, line 60: 5 character passwords? Is this more of a kind of PIN?

quirks/change-password-URLs.json
m33x: It's a forgot password link, where you can request a new password.

quirks/change-password-URLs.json
m33x: It's a forgot password link, where you can request a new password.

DerJuulsn:
Well, right they are forgot my password sites.

It's a password, but some call it still PIN. However they say it's safe because if you use digits, lower- and upper-case letters there are a few million possibilities, and that's enough. Stupid banks.

quirks/password-rules.json
m33x: That's normal with German banks, you have a randomly chosen password that is rather short e.g. 5 chars long. You can call it a PIN if you like. For example Sparkasse (one of the biggest German banks) calls it "PIN" in the login form. However, it's the secret that you need to sign into your bank account.

DerJuulsn:
Since its German it doesn't make that much sense for you, I guess. 
But I'm gonna translate it for you:

Consors Bank and Deutsche Bank: 

5 chars, use: letters and digits

HypoVereinsbank:

6 to 10 chars, use lower- and upper case letters and digits
you can also use: !"#$%&()*+:;<=>?@{}~

Commerzbank:

5 to 8 chars, use: letters and digits

Consors Finanz:

6 to 15 chars, you can use: letters, digits and "." "-"

Santander:

8 to 12 chars, use lower- and upper case letters and digits
you can also use: -!#$%&'()*,.:;=?^{}

rmondello:
This makes more sense now; thanks for the explanations!

Hey @m33x, you wanna try reviewing this one?

PR #97, opened 2020-06-06
Add linode.com to change password json

mike1e:
Adding linode.com to `change-password-URL.json`

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -9,7 +9,7 @@
     "geocaching.com": "https://www.geocaching.com/account/settings/changepassword",
     "gov.br": "https://acesso.gov.br/area-cidadao/#/alterarSenha",
     "impots.gouv.fr": "https://cfspart.impots.gouv.fr/monprofil-webapp/GererMonProfil",
-    "linode.com": "https://login.linode.com/forgot/password",
+    "linode.com": "https://cloud.linode.com/profile/auth",
     "microsoft.com": "https://account.live.com/password/Change",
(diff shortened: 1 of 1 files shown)


quirks/change-password-URLs.json
```diff
@@ -9,6 +9,7 @@
     "geocaching.com": "https://www.geocaching.com/account/settings/changepassword",
     "gov.br": "https://acesso.gov.br/area-cidadao/#/alterarSenha",
     "impots.gouv.fr": "https://cfspart.impots.gouv.fr/monprofil-webapp/GererMonProfil",
+    "linode.com": "https://login.linode.com/forgot/password",
```
rmondello, line 12: This looks like a forgot password URL, not a change password URL. Is that the case?

mike1e: @rmondello yeah seems like if you want to change the password from the settings, it takes you to the forgot password.

rmondello: Oh, cool. I can confirm that’s how Linode works (love Linode!). I wonder if it might be a better user experience to be taken to `https://cloud.linode.com/profile/auth` instead, because there they get an explanation of how they’ll have to reset their password to change their password. I don’t feel strongly about this; what do you think?

mike1e: I don't feel too strongly about it either. And looking at the other URLs, they get taken to the security page too. I'll update the code.

mike1e:
@rmondello Updated to point to the auth page in linode.

rmondello:
Thanks @mike1e!

PR #102, opened 2020-06-06
Add support for PilotFlyingJ

sicklepilot:
### Checklist
- [x] There are no [open pull requests](https://github.com/apple/password-manager-resources/pull) for the same update.


rmondello:
Hi @sicklepilot! I’d be happy to review this, but could you provide a screenshot or some copied text of the websites password requirements, or state that you derived them through experimentation? Thanks!

sicklepilot:
Sure thing @rmondello ,

I actually work on the mobile application team for this site and these are the requirements we have been given. Screenshot is attached. 

EF2D5880-66BF-408C-9227-6505B8B1C19F

rmondello:
PR looks good, although I think there might be a better path for you!

You can have your website directly adopt the Change Password URL:
https://wicg.github.io/change-password-url/

And Password Rules:
https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules

Being on these lists isn’t exactly a Good Thing! It basically say that the service in questions hasn’t adopted these two technologies. Maybe instead of landing this PR, you’d want to adopt these?

sicklepilot:
For sure, I’ve added a ticket internally for iOS but it might be some time before we can get it prioritized since it’s in an unofficial draft state. I’ll work with the website team as well. 

Furthermore these rules might potentially help android devices users as well as I’m not currently aware of a rule to set on that app for generation, but a password manager could still pull these JSON files in.

rmondello:
Your justification makes sense to me!

rmondello:
Thanks!

PR #103, opened 2020-06-06
Add consumidor.gov.br

rafaelvannucci:
### Checklist
- [X] There are no [open pull requests](https://github.com/apple/password-manager-resources/pull) for the same update.


quirks/change-password-URLs.json
```diff
@@ -6,6 +6,7 @@
     "ea.com": "https://myaccount.ea.com/cp-ui/security/index",
     "fnac.com": "https://secure.fnac.com/account/update-password",
     "geocaching.com": "https://www.geocaching.com/account/settings/changepassword",
+    "github.com": "https://github.com/settings/security",
```
rmondello, line 10: Github has adopted the Well-Known URL for Changing Passwords, so there’s no need to add it:
http://github.com/.well-known/change-password

rafaelvannucci: thanks

rmondello:
Thanks!

PR #106, opened 2020-06-06
Added atlassian.com (bitbucket.org) change password URL

m33x:
### Checklist


rmondello:
Thanks!

PR #107, opened 2020-06-06
Add Steam store password rules

sunbect:
### Checklist

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -198,6 +198,9 @@
     "ssa.gov": {
         "password-rules": "required: lower; required: upper; required: digit; required: [~!@#$%^&*];"
     },
+    "store.steampowered.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -198,6 +198,9 @@
     "ssa.gov": {
         "password-rules": "required: lower; required: upper; required: digit; required: [~!@#$%^&*];"
     },
+    "store.steampowered.com": {
+        "password-rules": " minlength: 6; required: lower; required: upper; required: digit;"
```
rmondello, line 202: Can you remove the leading space?

And no other characters are allowed?

sunbect: Special characters (~!@#$%^&*) are allowed but not required

rmondello: Then this might be a better rule:

```
minlength: 6; required: lower; required: upper; required: digit; allowed: [~!@#$%^&*];
```

PR #111, opened 2020-06-06
add shared backend websites for America's Test Kitchen

geoffcline:
America's Test Kitchen has 4 different recipe websites (with different target audiences), but all the sites share a login for saving recipes. 

I added the four domains to the list of websites with shared credential backends.

### Checklist

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -248,11 +248,5 @@
         "cooksillustrated.com",
         "cookscountry.com",
         "onlinecookingschool.com"
-    ],
-    [
-        "americastestkitchen.com",
-        "cooksillustrated.com",
-        "cookscountry.com",
-        "onlinecookingschool.com"
     ]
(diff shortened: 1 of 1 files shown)


quirks/websites-with-shared-credential-backends.json
```diff
@@ -242,5 +242,17 @@
...
+        "americastestkitchen.com",
+        "cooksillustrated.com",
+        "cookscountry.com",
+        "onlinecookingschool.com"
+    ],
+    [
+        "americastestkitchen.com",
+        "cooksillustrated.com",
+        "cookscountry.com",
+        "onlinecookingschool.com"
```
rmondello, line 256: It looks like you’ve added the group twice! Can you fix that up?

geoffcline: thank you -- i have added another commit to correc this

rmondello:
I think we need to figure out how sorting should work for this file, but not right now.

PR #113, opened 2020-06-06
add univ. of missouri domains w/ shared backend

geoffcline:
### Checklist


rmondello:
And there are login pages on each of these websites that all can be signed into with the same credentials?

geoffcline:
yes, they share a common active directory domain. 

for example, I can login to these sites with the same login
https://login.libproxy.mst.edu/login
https://ill.mul.missouri.edu/
https://www.umkc.edu/is/support/services/purchasing/software/Adobe.asp

rmondello:
And this is true for all five?

It also looks like there’s a conflict. Can you resolve that?

geoffcline:
Yes, it is true for all five. 

here are examples for the remaining 2 domains
https://sharepoint.umsystem.edu/
https://myview.umsl.edu/psp/csprds/?cmd=login&languageCd=ENG&

The same username/password/profile works on all 5 examples.

rmondello:
Sounds good. Looks like the current PR merges them into another group, rather than adding a new group.

geoffcline:
learning git one step at a time B) 

thanks! -- fixed in 5a0e81a (split the group)

rmondello:
I look forward to the University of Michigan one day adopting an SSO solution.

rmondello:
Thanks for putting up with all of the back and forth on this one, @geoffcline!

geoffcline:
no problem! thank you!

and for the record, they are working on a SSO solution (https://shib-idp.umsystem.edu/)

it's been in the works for at least 8 years. new apps use it, but the old things (peoplesoft campus, peoplesoft financial, library services), have huge userbases and are also proving difficult to update.

PR #114, opened 2020-06-06
add change password URL for univ. of missouri

geoffcline:
### Checklist

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -24,5 +24,6 @@
     "twitch.tv": "https://www.twitch.tv/settings/security",
     "xfinity.com": "https://customer.xfinity.com/users/me/update-password",
     "yahoo.com": "https://login.yahoo.com/account/change-password",
-    "zoom.us": "https://zoom.us/profile#pwd-form"
+    "zoom.us": "https://zoom.us/profile#pwd-form",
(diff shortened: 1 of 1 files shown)


quirks/change-password-URLs.json
```diff
@@ -24,5 +24,6 @@
     "twitch.tv": "https://www.twitch.tv/settings/security",
     "xfinity.com": "https://customer.xfinity.com/users/me/update-password",
     "yahoo.com": "https://login.yahoo.com/account/change-password",
-    "zoom.us": "https://zoom.us/profile#pwd-form"
+    "zoom.us": "https://zoom.us/profile#pwd-form",
+    "umsystem.edu": "https://password.umsystem.edu/reset/"
```
rmondello, line 28: This looks like a reset page. Is there a “change” page, or are changes done via reset?

geoffcline: changes are done via reset, if you visit the page it prompts you for user/pass, but provides options if you don't know the password. 

they require some kind of 2FA to change password (which is not generally required), which is historically why it is called reset

rmondello:
I’m satisfied with that explanation!

m33x:
But what about the alphabetic sorting :-/

rmondello:
I forgot my ABCs! Fixed in da63dde01624173c8fbff0515bab6e69e0a9600c.

PR #123, opened 2020-06-06
Added change password URL for Telekom.de (Telekom.com?)

m33x:
That's a difficult one, the German telekom.de and its news website t-online.de authenticate via telekom.com. I don't have a telekom.com (US) account, thus, I cannot verify whether this link also works for telekom.com, thus, I opted to add it as "telekom.de".

### Checklist


rmondello:
Oh, this one is a bit trickier. I’ll think about this a bit.

DerJuulsn:
Telekom.com isn't American. It's German.
The American website is t-mobile.com

DerJuulsn:
Bildschirmfoto 2020-06-06 um 23 43 48

rmondello:
If the authentication is done via telekom.com, would a user even have a saved password for telekom.de in the first place?

DerJuulsn:
I've just realized, there isn't even a possibility to log-in on telekom.com just telekom.de
That solves the problem, doesn't it? :)

rmondello:
Seems like. What do you think, @m33x?

m33x:
Yes, correct, so using 'telekom.de' as identifier was the correct choice. I suggest you allow the merge.

DerJuulsn:
Agree.

rmondello:
Thanks @m33x!

PR #128, opened 2020-06-06
Create websites-that-append-2fa-to-password.json

chrisjshull:
Start a list of websites which use a 2fa scheme which requires users to append the 2fa token to the password field.

Should be used to:
1) prevent auto-submission of login forms to allow the 2fa token to be added by the user
2) suppress "remember new password" prompts when the submitted password starts with the already stored password.

- USAA: https://www.usaa.com/inet/pages/security_token_logon_options and look under the "CyberCode Token" tab.
- E*TRADE: https://us.etrade.com/security-center/securityid

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ The _Password Manager Resources_ project exists so creators of password managers
 * **Password Rules**: Rules to generate compatible passwords with websites' particular requirements.
 * **Websites with Shared Credential Backends**: Groups of websites known to use the same credential backend, which can be used to enhance suggested credentials to sign in to websites.
 * **Change Password URLs**: To drive the adoption of strong passwords, it's useful to be able to take users directly to websites' change password pages.
+* **Websites that Append 2nd Factor to Password**: Some websites use a two-factor authentication scheme where the user must append a generated code to their password when signing in.
 
 Having password managers collaborate on these resources has three high-level benefits:
 
@@ -43,6 +44,13 @@ There are existing proposals to allow different domains to declare an affiliatio
 ### Change Password URLs
 
 The file [`quirks/change-password-URLs.json`](quirks/change-password-URLs.json) contains a JSON object mapping domains to URLs where users can change their password. This is the quirks version of the [Well Known URL for Changing Passwords](https://wicg.github.io/change-password-url/index.html). If a website adopts the Change Password URL, it should be removed from this list.
+
+### Websites that Append 2nd Factor to Password
+
+The file [`quirks/websites-that-append-2fa-to-password.json`](quirks/websites-that-append-2fa-to-password.json) contains a JSON array of domains which use a two-factor authentication scheme where the user must append a generated code to their password when signing in. Should be used to:
+
+1. Prevent auto-submission of login forms to allow the 2FA token to be added by the user.
+1. Suppress "remember new password" prompts when the submitted password starts with the already stored password.
 
 ## How to Contribute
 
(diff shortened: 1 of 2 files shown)


rmondello:
This is a great idea. Safari itself has a small list of such websites! I think this is a perfect fit for the project.

Would you mind writing up some changes to the README file to go along with this as part of the PR? You can use the structure of the other quirk types as a model.

chrisjshull:
README updated :)

README.md
```diff
@@ -9,6 +9,7 @@ The _Password Manager Resources_ project exists so creators of password managers
 * **Password Rules**: Rules to generate compatible passwords with websites' particular requirements.
 * **Websites with Shared Credential Backends**: Groups of websites known to use the same credential backend, which can be used to enhance suggested credentials to sign in to websites.
 * **Change Password URLs**: To drive the adoption of strong passwords, it's useful to be able to take users directly to websites' change password pages.
+* **Websites that Append 2nd Factor to Password**: Some websites use a two-factor authentication scheme where the user must append a generated code to their password when signing in.
```
rmondello, line 12: I’m going to get very particular here. What you have is good, but I’m going to try to help you make it even more clear.

I think this could be phrased slightly better. Websites aren’t doing the appending (implied by that). And maybe we can make this even more clear by referring to “2FA codes”, or something like that. How about?

Websites Where 2FA Code is Appended to Password

README.md
```diff
@@ -44,6 +45,13 @@ There are existing proposals to allow different domains to declare an affiliatio
 
 The file [`quirks/change-password-URLs.json`](quirks/change-password-URLs.json) contains a JSON object mapping domains to URLs where users can change their password. This is the quirks version of the [Well Known URL for Changing Passwords](https://wicg.github.io/change-password-url/index.html). If a website adopts the Change Password URL, it should be removed from this list.
 
+### Websites that Append 2nd Factor to Password
+
+The file [`quirks/websites-that-append-2fa-to-password.json`](quirks/websites-that-append-2fa-to-password.json) contains a JSON array of domains which use a two-factor authentication scheme where the user must append a generated code to their password when signing in. Should be used to:
```
rmondello, line 50: “Should be used to:” is kind of terse. How about, “This could be used to:”?

README.md
```diff
@@ -44,6 +45,13 @@ There are existing proposals to allow different domains to declare an affiliatio
...
+
+1. Prevent auto-submission of login forms to allow the 2FA token to be added by the user.
```
rmondello, line 52: I suggest, "Prevent auto-submission of login forms, allowing the user to append the 2FA code without frustration."

README.md
```diff
@@ -44,6 +45,13 @@ There are existing proposals to allow different domains to declare an affiliatio
...
+1. Suppress "remember new password" prompts when the submitted password starts with the already stored password.
```
rmondello, line 53: I suggest, “Suppress prompting to update a saved password when the submitted password has a prefix of the already-stored password."

rmondello: And, looking at this now, I bet these two points could be made as a single sentence or two sentences, without the load of a list.

My attempt:

```
The file [`quirks/websites-that-append-2fa-to-password.json`](quirks/websites-that-append-2fa-to-password.json) contains a JSON array of domains which use a two-factor authentication scheme where the user must append a generated code to their password when signing in. This list of websites could be used to prevent auto-submission of login forms, allowing the user to append the 2FA code without frustration. It can also be used to suppress prompting to update a saved password when the submitted password is prefixed by the already-stored password.
```

README.md
```diff
@@ -80,6 +88,10 @@ When contributing or amending a set of websites sharing a credential backend, yo
 
 Use the website in question until you find the standalone page for updating the user's password, or a high-level "Account Information" or "Security" page. The closer the URL takes the user to be able to change their password, the better. Before adding a URL, ensure that it works properly both when the user is logged in and when they are not. URLs added to [`quirks/change-password-URLs.json`](quirks/change-password-URLs.json) should have a scheme of https unless the website does not allow changing the password on an https page.
 
+### Contributing a Website that Appends 2nd Factor to Password
+
+When contributing or amending a set of websites that require that the user append a generated code to their password when signing in, you should state why you believe the relevant domains require such. This may invlove citing a URL to the relevant support page for the website.
```
rmondello, line 93: typo: invlove

README.md
chrisjshull: done

chrisjshull:
Suggestions applied :)

rmondello:
This looks good to me. Since it sets up a new type of quirk, I’m going to leave the PR open for a bit to gather additional feedback. I’ll try to merge it around Monday or Tuesday. :)

geoffcline:
I think this is noteworthy. 

Duo supports this (https://guide.duo.com/append-mode), and I've seen it used.

I'm also personally familiar with using USAA.

rmondello:
Merged! Again, great idea @chrisjshull!

PR #129, opened 2020-06-07
update PR template with def of shared backend

geoffcline:
based on the discussion in #58 I propose revising the pull request template.

This version clarifies when a group of domains is necessary.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,10 @@ An implementation of a parser for the Password Rules language that's written in 
 
 The file [`quirks/websites-with-shared-credential-backends.json`](quirks/websites-with-shared-credential-backends.json) contains a list of groups of websites that are known to share the same credential backend. This can be used to offer contextually relevant accounts to users on website A, even if credentials were previously saved for website B.
 
+In other words, a qualifying group serves login pages on each of the included domains. Those login pages accept accounts from the others. For example, `firstcompany.website/login` and `secondcompany.website/login` share the same backend (e.g., usernames, passwords, profiles), despite the different base domains.
+
+We don’t associate `google.com.li` to `google.com`, because `google.com.li` redirects to `accounts.google.com` for authentication.
+
 This list should not be used as part of any user experience that releases user credentials to a website without the user's explicit review and consent. In general, saved credentials should only be suggested to users with site-bound scoping. This list is appropriate for allowing a credential saved for website A to appear on website B if the website the credential was saved for is clearly stated.
 
 There are existing proposals to allow different domains to declare an affiliation with each other, which could be a way for websites to solve this problem themselves, given browser and password manager adoption of such a proposal. Until and perhaps beyond then, it is useful to have these groupings of websites to make password filling suggestions more useful.
--- a/README.md
+++ b/README.md
@@ -34,11 +34,9 @@ An implementation of a parser for the Password Rules language that's written in 
 
 ### Websites with Shared Credential Backends
 
-The file [`quirks/websites-with-shared-credential-backends.json`](quirks/websites-with-shared-credential-backends.json) contains a list of groups of websites that are known to share the same credential backend. This can be used to offer contextually relevant accounts to users on website A, even if credentials were previously saved for website B.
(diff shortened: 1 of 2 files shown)


igor-makarov:
@geoffcline perhaps the line should also give an example of what doesn't qualify? E.g. `google.co.il` redirecting to `accounts.google.com`.

Additionally, could you also split out the 'well-known' checklist item to a subheading titled "for password change URLs"?

geoffcline:
In my humble opinion, anything beyond a short example should be placed on a reference page. Perhaps I could also change the readme to have a section with this info, and link to it from the PR template?

I agree with splitting out the well known item.

igor-makarov:
People frequently don't read the readme files very thoroughly 😂

I think a short, concise example would be great to put in the PR template - something along the lines of "e.g. `google.co.il` redirecting to `accounts.google.com` doesn't need an entry".

The goal of the PR template is to save time for the primary maintainers, at the expense of a minor inconvenience to individual contributors.

rmondello:
This is good. My attempt to synthesize everything said here into one concise line:

The new group has login pages on each of the included domains, and those login page accept accounts from the others. (If example1.website redirects to example2.website for login, there is no need to add an association.)

igor-makarov:
I would still insist on adding a familiar example, but I like the phrasing for the Readme.

rmondello:
The new group serves login pages on each of the included domains, and those login page accept accounts from the others. (For example, we don’t associate google.com.li to google.com, because google.com.li redirects to accounts.google.com for authentication.)

geoffcline:
ok i did the following things 

I revised the shared backend checkbox based on feedback from @rmondello and @igor-makarov. I added a section to the readme with a positive and negative example for shared backends. 

I also added another checklist section for change password URLs. 

looking forward to more feedback

README.md
```diff
@@ -36,6 +36,10 @@ An implementation of a parser for the Password Rules language that's written in
 
 The file [`quirks/websites-with-shared-credential-backends.json`](quirks/websites-with-shared-credential-backends.json) contains a list of groups of websites that are known to share the same credential backend. This can be used to offer contextually relevant accounts to users on website A, even if credentials were previously saved for website B.
```
rmondello, line 37: I think we can achieve your goal here by amending this paragraph. Here’s my attempt, incorporating some of your great phrasing here.

> The file [`quirks/websites-with-shared-credential-backends.json`](quirks/websites-with-shared-credential-backends.json) contains a list of groups of websites that share the same credential backend and serve pages where users can sign in, accepting accounts from the others. For example, adding `first.website` and `second.website` means that `first.website` and `second.website` each serve a page (e.g. `first.website/login` and `second.website/login`) where the same accounts are valid for signing in, despite the different domains. It wouldn’t be appropriate to associate `google.com.li` to `google.com` because `google.com.li` redirects to `accounts.google.com` for sign-in, and `google.com.li` never serves a login page.
> 
> This data can be used by password managers to offer contextually relevant accounts to users on first.website, even if credentials were previously saved for second.website. This list should not be used as part of any user experience that releases...

README.md
```diff
@@ -36,6 +36,10 @@ An implementation of a parser for the Password Rules language that's written in
...
 
+In other words, a qualifying group serves login pages on each of the included domains. Those login pages accept accounts from the others. For example, `firstcompany.website/login` and `secondcompany.website/login` share the same backend (e.g., usernames, passwords, profiles), despite the different base domains.
+
+We don’t associate `google.com.li` to `google.com`, because `google.com.li` redirects to `accounts.google.com` for authentication.
```
igor-makarov, line 41: typo: `.il` not `.li`

@geoffcline

geoffcline: thanks!

README.md
geoffcline: thanks! 🤝

geoffcline:
Thanks for your feedback @rmondello -- I revised the description in the readme

rmondello:
This is fantastic. Thank you!

PR #130, opened 2020-06-07
Add a GitHub action to lint JSONs

igor-makarov:
Currently only checks sort order and JSON validity, will be extended in the future.

The file `quirks/websites-with-shared-credential-backends.json` is assumed to be ordered by the first element in each array.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  websites-shared-credentials:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+    # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
+    # change this to (see https://github.com/ruby/setup-ruby#versioning):
+    # uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
```
rmondello, line 28: Do we want to follow the comment’s advice?

.github/workflows/lint-scripts/websites-shared-credentials-sort-order.rb
```diff
@@ -0,0 +1,14 @@
+require 'json'
+
+shared_websites = JSON.parse(File.read('quirks/websites-with-shared-credential-backends.json'))
+
+shared_websites_sorted = shared_websites.sort_by(&:first)
+
+File.open('quirks/websites-with-shared-credential-backends.json','w') do |f|
+  f.write(JSON.pretty_generate(shared_websites_sorted))
```
rmondello, line 8: Can we maintain an indentation level of four spaces?

igor-makarov:
I'm gonna need to check about the ruby version.

In the meantime, I've experimented with validating password change URLs and some apparently work in the browser, but return 401 or 403. Any ideas?

rmondello:
> I'm gonna need to check about the ruby version.

Sounds good. Thanks for putting time into this. I think it’s going to be really cool.
> In the meantime, I've experimented with validating password change URLs and some apparently work in the browser, but return 401 or 403. Any ideas? Can we even validate without a valid login?

I think making a network request is a validation step that goes too far, because a transient network condition could cause failure. In practice, it could be maddening. Perhaps the furthest we could go for validating the change password URLs if the file being indented properly, the keys and values being strings, and the values looking like URLs.

igor-makarov:
@rmondello done!

igor-makarov:
whoops, not done @rmondello

.github/workflows/lint-scripts/websites-shared-credentials-sort-order.rb
igor-makarov: Ruby defaults to 2 spaces - https://stackoverflow.com/questions/6741586/rails-coding-standards-why-2-space-indentation

Do you still want 4?

igor-makarov:
done the rebase and fixed the ruby action tag

.github/workflows/lint-scripts/websites-shared-credentials-sort-order.rb
rmondello: I’m sorry! I meant for the data itself, not your action’s code. I think `JSON.pretty_generate` can take an indentation level.

igor-makarov: Added 4-space indent to JSON formatter.

igor-makarov:
@rmondello I've added another lint, that validates that all JSONs are valid.

Promise not to push more changes 🤞

rmondello:
I’m working to enable actions on the repository before landing this.

igor-makarov:
@rmondello is there any specific trouble enabling actions? I thought the feature is out of beta now...

darinadler:
Can we use the script to do a one-time sort of these files while we wait for GitHub Actions to be enabled?

darinadler:
I suspect there may be a bit of delay in enabling actions because it might need to be done at the same time across all Apple GitHub projects?

igor-makarov:
@darinadler it's possible to use the script for local sorting, but it'll just get unsorted again without linting...

This way, I'll be able to make sure that it's well-sorted when the PR lands.

darinadler:
I think there’s significant value on having it mostly-correctly-sorted before your script lands.

rmondello:
I’m working with some folks to see if we can enable Github actions. In the meantime, I’m OK with merging this PR so that @igor-makarov can enable them on their fork and continue to iterate. Does that sound OK to you, @igor-makarov?

igor-makarov:
I have no problem with that!

igor-makarov:
Ok I just rebased and several things in master branch was out of alphabetical order. If the action won't run, it will get unordered pretty soon...

igor-makarov:
It's mergeable though. @rmondello

geoffcline:
Could we implement this as a script users run locally?

Or as a test that returns a warning for commits w/ not properly sorted data?

Just some thoughts.

rmondello:
Merging this, although it won’t be active until later.

PR #133, opened 2020-06-07
Add udel edu

willswire:
### Checklist


quirks/change-password-URLs.json
```diff
@@ -49,6 +49,7 @@
     "tripit.com": "https://tripit.com/account/edit/section/change_password",
     "twilio.com": "https://www.twilio.com/console/user/settings",
     "twitch.tv": "https://www.twitch.tv/settings/security",
+    "udel.edu": "https://udapps.nss.udel.edu/passwordReset/selfservice",
```
rmondello, line 52: Are all change password requests for udel.edu password resets?

quirks/password-rules.json
```diff
@@ -216,6 +216,9 @@
     "telekom-dienste.de": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [#$%&()*+,./<=>?@_{|}~];"
     },
+    "udel.edu": {
+        "password-rules": "minlength: 12; maxlength: 30; required: lower; required: upper; required: digit; required: [!@#$%^&*()+];"
```
rmondello, line 220: Looks good!

quirks/websites-with-shared-credential-backends.json
```diff
@@ -248,5 +248,9 @@
         "cooksillustrated.com",
         "cookscountry.com",
         "onlinecookingschool.com"
+    ],
+    [
+        "udel.edu",
+        "microsoftonline.com"
```
rmondello, line 254: Can you explain this? Are there login pages that both udel.edu and microsoftonline.com that accept the same credentials? This seems a little off to me.

willswire: The second URL should be https://www.office.com/. This is because Microsoft Office 365 Accounts with the same credentials as UDel Net IDs are automatically provisioned in the back-end.

quirks/password-rules.json
willswire: great!

rmondello:
I removed the credential backend association from this PR. I’m a little nervous about the association between “udel.edu” and “office.com”, and I think it requires a bit more thought. But I didn’t want to block your other changes from getting in. @willswire, can you open a new PR for the association, so that we can focus on it there? Thank you so much!

willswire:
Not a problem, @rmondello! I do need to change one commit (I was working on it but you merged the PR).  The change password link should be: https://udapps.nss.udel.edu/myUDsettings/password

PR #137, opened 2020-06-07
Add alliantcreditunion.com and sort alphabetically

parkerbxyz:
This PR adds password rules and the change password URL for alliantcreditunion.com. It also fixes the alphabetical order of existing JSON objects.

### Checklist


rmondello:
Thanks @parkerbxyz!

PR #142, opened 2020-06-07
Add zocdoc.com's Change Password URL

davidquesada:
If you visit this URL while not signed in, you are taken to a login page
that redirects to the Change Password page after successful authentication.

### Checklist

- [ ] The new group serves login pages on each of the included domains, and those login page accept accounts from the others. (For example, we don’t associate google.com.li to google.com, because google.com.li redirects to accounts.google.com for authentication.)


rmondello:
Thanks for reviewing, @bradcush and @DerJuulsn!

PR #154, opened 2020-06-08
Add joyn.de

janek26:
### Checklist

- [x] The new group serves login pages on each of the included domains, and those login page accept accounts from the others. (For example, we don’t associate google.com.li to google.com, because google.com.li redirects to accounts.google.com for authentication.)


rmondello:
Can you give me a little bit more context about joyn.at? It looks like it’s not fully up and running right now?

janek26:
true, it‘s not right now. Should I remove it til then?

igor-makarov:
Are you the developer for the websites? Because if you are, there's a better way to support password change, and it's preferred:  
`https://example.com/.well-known/change-password`

This way, you do not need to be on the password change list.

janek26:
Alright, I can adopt that. 
Can we keep the `websites-with-shared-credential-backends.json` modification?

igor-makarov:
Yes, I think there's no decentralized alternative for this. Am I right @rmondello?

rmondello:
@janek26 Are you going to offer login web pages on both websites that use the same credential backend? That’s the only legitimate reason to be added to this list. But really, the best practice is for you to use a more SSO approach, where login is centralized to a single domain.

janek26:
Alright, thanks for the feedback 👍

rmondello:
I super appreciate your enthusiasm for this! And I’m happy that in these cases, we don’t need to add quirks for your website, because you can go and do the right things! :-)

PR #151, opened 2020-06-08
Check duplicates in the shared backend list

igor-makarov:
This PR verifies that `websites-with-shared-credential-backends.json` doesn't contain duplicate domains.

PR #156, opened 2020-06-08
Sort websites with shared credential backends

darinadler:
### Checklist


quirks/websites-with-shared-credential-backends.json
```diff
@@ -1,273 +1,274 @@
...
+        "kp.org"
     ],
     [
-        "alltrails.com",
-        "alltrails.io"
+        "live.com",
+        "microsoft.com",
+        "microsoftonline.com",
+        "office.com",
+        "skype.com"
```
darinadler, line 206: This used to be two separate lists, both including live.com. This merges them.

igor-makarov:
Thanks for the initiative!
However, #130 covers the sorting by using a script. #151 covers the duplicate checks. 

Consider closing the PR in favor of those?

darinadler:
I’ll certainly drop this if those get merged!

igor-makarov:
Yeah, they got held up because GH Actions aren't enabled for this repo yet...

Issue #144, opened 2020-06-08
shared backend appropriate when one domain is a subdomain?


geoffcline:
Is a shared backend group appropriate when one domain is a subdomain?

For example, you enter your seattle public library username/pin on pages served from both these sites:
seattle.bibliocommons.com (library catalog)
spl.overdrive.com (ebooks)
spl.org (access online library resources, eg, databases)

Is the fact this rule only works for "seattle.bibliocommons.com" a problem? 

Consider, the St. Louis Public Library uses:
slpl.bibliocommons.com
slpl.overdrive.com
slpl.org

igor-makarov:
The three domains you listed aren't considered subdomains of each other, therefore listing them as shared credentials is a good idea. However, it appears on first glance that `spl.org` redirects to one of the other domains for login, so it's not necessary to add this one.

geoffcline:
Ok, thanks for this feedback, I will add two different groups for "seattle.bibliocommons.com" and "slpl.bibliocommons.com" 

small note, seattle uses ezproxy.spl.org/login for access to some resources

PR #157, opened 2020-06-08
add kansas city, st louis, and seattle library groups to shared backends

geoffcline:
based on the discussion in #144 -- i added the following library groups. 

They all include subdomains of [libraryname].overdrive.com

It's definitely confusing to keep track of the Seattle Public Library (SPL), St Louis Public Library (SLPL), St Louis County Library (SLCL), Kansas City Library (KCL), and King County Library System (KCLS). 

### Checklist
- [ ] The top-level JSON objects are sorted alphabetically 


geoffcline:
oh, and yes I did triple check that login pages are served for each of these domains B)

igor-makarov:
Wait, so different subdomains of `bibliocommons.com` would have different credential backends? 😭

geoffcline:
yes! it is a pain! i tried to say this in #144 but it appears it is still an issue ... let me know how I can clarify or help.

rmondello:
One thing that we’ve done with iCloud Keychain is put bibliocommons.com on a short list of websites that are known to be affiliated with so many different credential backends that it doesn’t make sense to warn users about password reuse with that website, because the odds of it being a legitimate manifestation of a shared credential backend situation are too high. We thought this list and purpose was pretty specific to our product, but this may not be the case.

rmondello:
Looks like we need some rebasing.

geoffcline:
I’ll do more thinking on the concept of a “sub domains with different backends” list compared to the current “websites with shared backends” list.

Libraries also have the interesting pattern of the credentials (usually) being an assigned barcode as the username and a four digit pin.

I think it’s worthwhile here — the assigned barcode is a pain to remember, and as we can see here you enter it on a variety of different sites.

rmondello:
@geoffcline I’d love to help you get this merged. Can you fix up the merge conflicts?

geoffcline:
Yes, I will fix the conflicts soon. Thank you!

rmondello:
I know that @geoffcline really wants to get this in, so leaving this open. (I’m doing a pass of open PRs right now.)

geoffcline:
no more conflicts ! ! ! 🥳

thanks for waiting.

PR #164, opened 2020-06-08
adds getflywheel.com password change urls and requirements

jacksellwood:
### Checklist


quirks/change-password-URLs.json
```diff
@@ -15,6 +15,7 @@
     "ea.com": "https://myaccount.ea.com/cp-ui/security/index",
     "fnac.com": "https://secure.fnac.com/account/update-password",
     "geocaching.com": "https://www.geocaching.com/account/settings/changepassword",
+    "getflywheel.com": "https://app.getflywheel.com/password_reset",
```
rmondello, line 18: Is there a change URL anywhere, or are all changes done by resetting?

PR #166, opened 2020-06-08
Add Disney+ to Disney/go.com shared backend

plttn:
Add Disney+

### Checklist

`whois disneyplus.com`
image
`whois go.com`
image


karlw00t:
Why was this closed? I had an issue with this recently and I think adding disenyplus.com to the list is the correct call.

plttn:
I PR'd from the wrong branch and forgot to fetch prior to committing, so the merge was funky.

Issue #98, opened 2020-06-06
no unit tests or comments


prabhu:
This is a good initiative. But there are:

- no unit tests
- no comments so not able to follow several flows. For eg: what is the method `_parsePasswordRequiredOrAllowedPropertyValue` doing?
- Single line if conditions are lacking braces. Did we forget [goto fail?](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/)

Hopefully the code will get improved over time.

rmondello:
> This is a good initiative.

Thanks!

> But there are:
> 
> * no unit tests

I have a test suite written for the parser, but I haven’t figured out how to make it or something like it run on Github yet. I’ll inline it at the end of this comment, and maybe you’ll have ideas on how we can integrate it into the project. It’s a Ruby script that has a dependency on `execjs`.

> * no comments so not able to follow several flows. For eg: what is the method `_parsePasswordRequiredOrAllowedPropertyValue` doing?

Fair! I personally might not be able to get around to patching this up soon, but PRs are welcome!

> * Single line if conditions are lacking braces. Did we forget [goto fail?](https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/)

This is a holdover from the WebKit project’s style. I’d agree with you that we should deviate from it here. If you’d like to send in a PR for that, I’d happily review it. Otherwise, I’ll get around to that sometime!

> 
> Hopefully the code will get improved over time.

```
#!/usr/bin/env ruby

PATH_TO_CONTAINING_DIRECTORY = File.expand_path(File.dirname(__FILE__))
$LOAD_PATH.unshift "#{PATH_TO_CONTAINING_DIRECTORY}/build-autofill-quirks-helpers/vendor/execjs-2.6.0/lib"

require 'execjs'

input_path = ARGV.shift

if !input_path
  STDERR.puts "Need to specify both an input parser path."
  exit 1
end

password_rules_parser_javascript = File.read input_path
javaScriptContext = ExecJS.compile password_rules_parser_javascript

DEFAULT_RULES = '[{"_name"=>"allowed", "value"=>[{"_name"=>"ascii-printable"}]}]'

input_and_expected_results = [
  ["", DEFAULT_RULES],
[...]

PR #168, opened 2020-06-08
Simple test framework using Mocha, initial tests for validating JSON files

billjive:
### Checklist

I thought it would be useful to add a basic testing framework to this project. It uses Mocha and is run locally with node/npm. I added a package.json file which has basic info but also dev dependencies (just Mocha and another library to validate URLs). To run, do "nom install" followed by "nom test" in the main directory.

Current tests just validate the basics in the JSON files. All entries are validate to be valid URLs, for example.

Future work could be to validate the password rules per the WHATWG spec.


rmondello:
Oh, neat. @igor-makarov, what do you think?

igor-makarov:
Looking good!
I have a few notes:
1. I'd have the tests for different quirk files in separate source files, for easier diffing/merging.
2. Some of the Ruby-based validations in #130 & #151 can be rewritten in JS so that the test suite stays homogenous in terms of language/framework.
3. Validating password rules can be done using the `tools/PasswordRulesParser.js` parser instead of regex, am I right @rmondello?
4. Let's add it to GitHub Actions!

igor-makarov:
@billjive if you rebase, I can add the GitHub-based workflow to your branch!

billjive:
> @billjive if you rebase, I can add the GitHub-based workflow to your branch!

Great! Thanks everyone for the feedback, good suggestions. I can get this going in a day or two, sound good?

billjive:
> @billjive if you rebase, I can add the GitHub-based workflow to your branch!

Ok -- all rebased. I also split up the quirks tests in to 3 (well, now 4) files. I added a sanity check test case on the new 2fa json file.

I looked at the other PR's and I can easily add a test case to check for duplicate entries, etc. Any other ideas?

igor-makarov:
@billjive I made a PR to your branch with the addition of the test action. See it run here: https://github.com/igor-makarov/password-manager-resources/actions

billjive:
Great! Merged..

rmondello:
@billjive This is really cool. It looks like it might need to be rebased. I was looking over the dependencies JSON file. That’s *a ton* of dependencies, no?

billjive:
@rmondello thanks -- it's been a while so yes I'm sure it needs to be rebased. I'll figure out too if there are too many dependencies, not sure at the moment but will look. I'll get to this stuff over the weekend I think.

I also rewrote your ruby test suite in javascript. It was a bigger rewrite and I can share with you in a different branch. Works just the same!

rmondello:
@billjive Very interested in seeing the Ruby->JavaScript re-write!

billjive:
@rmondello I posted details here: https://github.com/apple/password-manager-resources/issues/98#issuecomment-671488154

RubenMateus (approved):

package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "password-manager-resources",
```
jedmao, line 2: Should probably mark this as private, yeah?
```suggestion
    "name": "password-manager-resources",
    "private": true,
```

billjive: Sure, up to you guys.

tdung (approved):

rmondello:
I'm interested in evaluating this proposed set of changes, but they don't merge cleanly. In order to consider reviewing this, it will need to be rebased.

PR #175, opened 2020-06-09
add clien.net password rule and change password url

natz92:
After confirming the password on the confirm page, change password.

### Checklist


quirks/password-rules.json
```diff
@@ -59,6 +59,9 @@
     "claimlookup.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [@#$%^&+=!];"
     },
+    "clien.net": {
+        "password-roles": "minlength: 5; required: lower, upper; required: digit;"
```
jihoon-seo, line 63: `password-roles` should be corrected to `password-rules`.

natz92: > `password-roles` should be corrected to `password-rules`.
rebase done. thank you :)

PR #176, opened 2020-06-09
Add some Korean sites' change password URLs

jihoon-seo:
### Checklist


rmondello:
Hi @jihoon-seo! I fixed the merge conflict on this, and removed GitHub. Do you agree to the Developer Certificate of Origin? https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md

jihoon-seo:
@rmondello Yes, I agree!

quirks/change-password-URLs.json
```diff
@@ -65,6 +73,7 @@
     "minecraft.net": "https://www.minecraft.net/profile",
     "mojang.com": "https://www.minecraft.net/profile",
     "myaccount.ea.com": "https://myaccount.ea.com/cp-ui/security/index",
+    "naver.com": "https://nid.naver.com/user2/help/myInfo.nhn?m=viewChangePasswd&lang=ko_KR",
```
rmondello, line 76: Okay, one last bit. Should we remove the `lang` parameter? Does the link work without it?

jihoon-seo: Yes it seems that the link works without it, so it will be nice if the `lang` parameter is removed.

rmondello:
Thank you! :-)

PR #178, opened 2020-06-09
Add password rule for robinhood.com (min only required)

cljazouli:
### Checklist
- [X] There are no [open pull requests](https://github.com/apple/password-manager-resources/pulls) for the same update.

This is to add the password requirements for https://robinhood.com/ .
The only requirement is a min_length of 10.

image

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -213,6 +213,9 @@
     "riteaid.com": {
         "password-rules": "minlength: 8; maxlength: 15; required: lower; required: upper; required: digit;"
     },
+    "robinhood.com": {
+        "password-rules": "minlength: 10;"
+    },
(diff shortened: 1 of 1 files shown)


bradcush:
I'm in support of adding this as I think a min of 10 is outside the nominal case

cljazouli:
@rmondello Having nightmares about this never getting merged

rmondello:
@cljazouli Let’s fix those nightmares. Do you agree to the Developer Certificate of Origin? https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md

cljazouli:
I do ! 🚀  @rmondello

rmondello:
Thanks @cljazouli!

Issue #192, opened 2020-06-10
Add leetchi.com


jakub-g:
https://www.leetchi.com/en/User/PlainRegister

![image](https://user-images.githubusercontent.com/1437027/84291591-9b9be600-ab45-11ea-9b94-6c01998ac4b2.png)

domain: `www.leetchi.com`
rules:

8 characters
1 lower case letter
1 upper case letter
1 number
1 special character (@, #, &, etc.)

jakub-g:
done in #193

PR #193, opened 2020-06-10
Add leetchi.com

jakub-g:
See #192

### Checklist

- [x] There is no Well-Known URL for Changing Passwords (`https://www.leetchi.com/.well-known/change-password`)


jakub-g:
Hi @rmondello thx for checking - I marked it as draft as I only briefly checked all the special chars. In UI they seem to be accepted but I didn't verify if backend really accepts all of them.

crleonard:
This looks good to me, follows pattern and clean.

PR #200, opened 2020-06-11
Add Wikimedia projects

edwardwall:
### Checklist

These are all Wikimedia projects which share credentials https://www.wikimedia.org/

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -263,7 +263,6 @@
         "wayfair.ca"
     ],
     [
-        "wikimedia.org",
         "wikipedia.org",
         "mediawiki.org",
         "wikibooks.org",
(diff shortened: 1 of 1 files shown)


rmondello:
Hi @edwardwall! Can you do a little storytelling here? Are all accounts on all of these websites automatically valid on all of the others? Only some accounts? Is this still the case, or is there some amount of history at play here?

edwardwall:
There are some wikimedia.org subdomains which use [non-standard accounts](https://foundation.wikimedia.org/wiki/Privacy_policy/FAQ#standardaccount) and therefore don't share credentials.  
I've removed wikimedia.org and added specific subdomains. Every account created on any of those websites can be used on all the others.  


However there are quite a few other wikimedia.org subdomains which share credentials, eg bd.wikimedia.org and nyc.wikimedia.org. This is [the most complete list I could find](https://meta.wikimedia.org/wiki/Wikimedia_wikis).

PR #216, opened 2020-06-14
axa.de - password rule

jumoru:
**Adding password rule for axa.de**

- Minimum length = 8 (see screenshot), maximum length = 65 (determined by testing)
- Additionally the password has to contain 3 out of 4 of the following criteria: lower, upper, digit, special characters (Proposed password rule requires lower, upper and digits. Special characters are allowed.) 

Passwort mit mindestens 8 Zeichen

### Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -30,7 +30,7 @@
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!\"#$%&'()*+,./:;<=>?@[^_`{|}~]];"
     },
     "axa.de": {
-        "password-rules": "required: lower; required: upper; required: digit; allowed: [!\"§$%&/()=?;:_-+*\'#]; minlength: 8; maxlength: 65;"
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -29,6 +29,9 @@
     "autify.com": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!\"#$%&'()*+,./:;<=>?@[^_`{|}~]];"
     },
+    "axa.de": {
+        "password-rules": "required: lower; required: upper; required: digit; allowed: [!“§$%&/()=?;:_-+*’#]; minlength: 8; maxlength: 65;"
```
DerJuulsn, line 33: “ is the wrong one, use \\" instead

DerJuulsn: same for '

jumoru: Thank you very much for alerting me to this. It should be fixed now.

rmondello:
Thank you!

PR #219, opened 2020-06-15
Added Change Password URL

r1:
### Checklist


DerJuulsn:
Could you change the PR title to sth more meaningful? :)

PR #221, opened 2020-06-15
added rules for twitter

Rishikesh-12:
### Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -270,8 +270,8 @@
     "telekom-dienste.de": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [#$%&()*+,./<=>?@_{|}~];"
     },
-    "twitter.com":{
-        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [`~!@#$%^&*()-_+={}[]:;'\"<>,.?/|\\];"
+    "twitter.com": {
+        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [-`~!@#$%^&*()_+={}:;'\"<>,.?/|\\[]];"
     },
     "ubisoft.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [-]; required: [!@#$%^&*()+];"
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -271,7 +271,7 @@
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [#$%&()*+,./<=>?@_{|}~];"
     },
     "twitter.com": {
-        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [-`~!@#$%^&*()_+={}:;'\"<>,.?/|\\[]];"
+        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: special;"
     },
     "ubisoft.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [-]; required: [!@#$%^&*()+];"
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -271,7 +271,7 @@
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [#$%&()*+,./<=>?@_{|}~];"
     },
     "twitter.com":{
-        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [!@#$%^&*()+]"
+        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [`~!@#$%^&*()-_+={}[]:;'\"<>,.?/|\\];"
     },
     "ubisoft.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [-]; required: [!@#$%^&*()+];"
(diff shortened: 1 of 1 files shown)


Rishikesh-12:
added rules for twitter

rmondello:
Can you provide some primary sourcing information on how you created this requirement? It looks valid and fine, but it’s good to do this.

Rishikesh-12:
Yes, I searched for a url which wasn't already present in the password-rule.json then I searched for the password standards of twitter which resulted in minimum of 8 character long password.

rmondello:
This isn’t really good enough for us. Can you provide screenshots or links to Twitter’s documentation for their password requirements, or can you claim that you experimented to come up with the exact list of special characters that are allowed?

Rishikesh-12:
Here is a document i came across

About_account_security

rmondello:
Great! We’re almost there! Where did the maxlength come from?

Rishikesh-12:
Use_long_random_string_for_a_password_preferably_32_characters_long_say_experts_Technology_News_Firstpost

jumoru:
Are you sure that there really is an upper limit on how many characters can be used? I haven't tested all possibilities, but I was able to choose a password with more than 40 characters. Based on my quick tests I'm also unsure whether Twitter really **requires** all of these categories. For example, I was able to choose a password without special characters.

Rishikesh-12:
**Yes**, the only requirement by them is  _minimum 8 characters_ , I added extra requirements after keeping security in mind, I can create a new pr if you want, with **only** minimum character length requirement.

jumoru:
Got it, thank you for your response.

Based on my current understanding of how these rules work in the background (please correct me when I’m wrong), I think it actually would be more secure to just require at least 8 characters. This is because the algorithm then just randomly selects at least 8 characters out of the set of all possible characters (lower, upper, digit, special characters, ...), which gives us a higher number of possible password than when we require at least 4 characters to be of a certain subtype.

Having said that, there already was some discussion whether we need a password rule for such scenarios to begin with, because probably no decent password manager would choose a password with less than 8 characters.

Unfortunately I don’t know how this discussion has ended. Does anyone else here know how we deal with such cases?

rmondello:
It’s an ongoing discussion. This example will help us tune our (currently undocumented) parameters.

Here, I think that “at least eight characters; can contain anything” is a Good policy for Twitter to have, and password manager software won’t have to work around it. Therefore, I don’t see a reason to add a rule to the project here.

Rishikesh-12:
okay

rmondello:
@Rishikesh-12 Actually, this looks like a valuable rule to me, and I’d like to help you merge it. Can you check if Twitter allows the `-` character?

Also, do you agree to the project’s Developer Certificate of Origin? https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md

Rishikesh-12:
Yaa sure.

Rishikesh-12:
Check this out.

https://www.awesomescreenshot.com/video/392217?key=98966f7d34455af8e74ed3132f0e9a78

rmondello:
@Rishikesh-12 The rule in your PR doesn’t include `-` as allowed. I suspect there are more characters that they’d allow if you tested them.

Rishikesh-12:
Definitely I would check and get back to you.

Rishikesh-12:
okay so I have checked this for every symbol on my keyboard.

https://www.awesomescreenshot.com/video/483655?key=d3b8e7ac81c959aa792c36f0f9d5e1d0

Rishikesh-12:
Also I have updated the pull request.

quirks/password-rules.json
```diff
@@ -270,6 +270,9 @@
     "telekom-dienste.de": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [#$%&()*+,./<=>?@_{|}~];"
     },
+    "twitter.com": {
+        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [-`~!@#$%^&*()_+={}:;'\"<>,.?/|\\[]];"
```
rmondello, line 274: So it turns out that your required set is one character away from `special`, and the only character that’s missing is space (` `). Does Twitter allow spaces in passwords? If so, we should update this last part to `required: special;`.

Rishikesh-12: https://www.awesomescreenshot.com/video/487826?key=84e7ca7faf7de1c441aeac43410ff15e


Okay I have checked the condition and its working with 'space' and i have also updated the pr.

rmondello:
Based on the comments from @jumoru earlier, I think this rule might be closer to the truth:

```
minlength: 8; required: lower; required: upper; required: digit; allowed: special;
```

Which is equivalent to:
```
minlength: 8; required: lower; required: upper; required: digit; allowed: ascii-printable;
```

But I like the first one more.

Rishikesh-12:
So is there any changes, I need to make in the PR ?

jumoru:
Let me try to sum up what we have so far:

- @Rishikesh-12, if I've understood you correctly a few comments ago we both agree that Twitter only requires at least 8 characters (f.e. a password can contain just numbers)
- if I've understood @rmondello correctly we have now established that Twitter accepts all ascii printable characters

So, my question would be whether `minlength: 8; allowed: ascii-printable;` would be sufficient here? Thoughts?

Rishikesh-12:
@jumoru Correctly said ascii-printable would be sufficient !

* To support the claims i have attached a chart of ascii printable.
      https://en.wikipedia.org/wiki/ASCII#ASCII_printable_code_chart

jumoru:
Great, thanks for confirming this! I just took another look at the [documentation](https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules) of the password rule language. Given that ascii-printable seems to be the default I think we could simplify this even further to `minlenght: 8`

@rmondello, is this correct or do I miss something?

rmondello:
@jumoru I think that’s right.

Rishikesh-12:
So, should i update the pull request to minlength:8 only ?
@rmondello

rmondello:
> So, should i update the pull request to `minlength:8` only ?
> @rmondello

I think so.

Rishikesh-12:
Okay I have updated the pull request.

PR #224, opened 2020-06-15
 Add password rules for virginmobile.ca

HarshPatel270698:
### Checklist

#### for password-rules.json

image


rmondello:
Thanks @HarshPatel270698!

PR #225, opened 2020-06-16
In character classes that include the character '-', move '-' to be t…

rmondello:
…he first character, as required by the specification draft

### Checklist
- [X] There are no [open pull requests](https://github.com/apple/password-manager-resources/pulls) for the same update.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -3,7 +3,7 @@
         "password-rules": "minlength: 6; maxlength: 16;"
     },
     "aetna.com": {
-        "password-rules": "minlength: 8; maxlength: 20; required: upper; required: digit; max-consecutive: 2; allowed: lower, [_&-#@];"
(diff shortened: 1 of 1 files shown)


DerJuulsn:
@rmondello 
So, I'm gonna add this rule as well I guess 🤣
Are there any other such rules?

rmondello:
> So, I'm gonna add this rule as well I guess 🤣

Sorry — add it to where?

> Are there any other such rules?

In a character class, if present, `-` must be the first character, and if present, `]` must be the last character. `-`’s rule allows for a syntax expansion to include ranges in the future. `]`’s rule resolves an ambiguity about the end of a character class.

DerJuulsn:
@rmondello I'm sorry! To the web-interface.

Ah okay. So thats all, thanks for your response!

rmondello:
Ah, that makes sense! No need to be sorry, @DerJuulsn. I was just confused, and you helped me not be confused anymore. 😅

PR #238, opened 2020-06-18
add password rules for 1800flowers

bradcush:
1800flowers-password-rules

### Checklist
- [x]  I agree to the project's Developer Certificate of Origin.

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -2,6 +2,9 @@
     "163.com": {
         "password-rules": "minlength: 6; maxlength: 16;"
     },
+    "1800flowers.com": {
+        "password-rules": "minlength: 6; required: lower, upper; required: digit;"
+    },
     "aetna.com": {
         "password-rules": "minlength: 8; maxlength: 20; required: upper; required: digit; max-consecutive: 2; allowed: lower, [-_&#@];"
     },


rmondello:
Hi @bradcush,

We recently made a change to the project to clarify the source of contributions to the project. That involved us changing the pull request template to have this item:

- [ ] I agree to the project's [Developer Certificate of Origin](https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md)

Can you confirm that you agree to this?

bradcush:
@rmondello Yes, confirming here that I agree to the "Developer Certificate of Origin".

bradcush:
@jumoru Yes but as there is no way of specifying that either a lowercase or uppercase letter is required I think it's best to include both as required, no? I'm not sure there is a better way to specify that and I prefer having them both as required instead of only one set as required and the other allowed or optional. Do you have a better suggestion for this type of case?

DerJuulsn:
@bradcush AFAIK you can use required: lower, upper to say lower or upper, but at least one of them.

bradcush:
@DerJuulsn Ok cool, if that's the case I'll go ahead to make that change. I didn't realize there was a difference between specifying them as `required: lower, upper;` or `required: lower; required: upper;`. Thanks!

jumoru:
@bradcush @DerJuulsn This is what I was thinking about as well. It's documented [here](https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules).

DerJuulsn:
@jumoru thanks for the source. Yep, it's kinda hidden.

bradcush:
@DerJuulsn, @jumoru I've updated the PR accordingly

quirks/password-rules.json
```diff
@@ -2,6 +2,9 @@
     "163.com": {
         "password-rules": "minlength: 6; maxlength: 16;"
     },
+    "1800flowers.com": {
```
Cor109, line 5: Corleone

PR #273, opened 2020-06-28
added makemytrip.com

3NCRY9T3R:
### Overall Checklist

#### for password-rules.json


quirks/password-rules.json
```diff
@@ -216,6 +216,9 @@
     "macys.com": {
         "password-rules": "minlength: 7; maxlength: 16; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789~!@#$%^&*+`(){}[:;\"'<>?]];"
     },
+    "makemytrip.com": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [@$!%*#?&];"
```
rmondello, line 220: Can you provide some primary source information about how you derived this rule? It looks like a good rule, and I’m happy to merge, but we’d like to have this information on hand.

rmondello:
Thanks @3NCRY9T3R!

Issue #134, opened 2020-06-07
Document what a set of password requirements not worthy of incorporation into the project is


rmondello:
In https://github.com/apple/password-manager-resources/pull/86, David Jennes writes:

> This is maybe a more general discussion for this project: do we want to document every password quirk for each website? Or can we agree to a set of rules that we assume (or require?) that each password manager will at least adhere to? I'm pretty sure that each password manager will generate something longer than 8 characters. Not sure if the same can be said for 16 chars (or longer), that's something that may be dependent on the user's settings.

This seems worth doing, to me.

geoffcline:
I think microsoft's documentation for windows 10 is a relevant example of "default password requirements"

https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/password-must-meet-complexity-requirements#reference

In relevant part:

> The password contains characters from three of the following categories:
> - Uppercase letters of European languages (A through Z, with diacritic marks, Greek and Cyrillic characters)
> - Lowercase letters of European languages (a through z, sharp-s, with diacritic marks, Greek and Cyrillic characters)
> - Base 10 digits (0 through 9)
> - Non-alphanumeric characters (special characters): (~!@#$%^&*_-+=`|\(){}[]:;"'<>,.?/) Currency symbols such as the Euro or British Pound are not counted as special characters for this policy setting.
> - Any Unicode character that is categorized as an alphabetic character but is not uppercase or lowercase. This includes Unicode characters from Asian languages.

The Azure AD (msft's cloud directory product) documentation isn't very helpful. It only talks about min length and a generic "is complex enough" standard.

https://docs.microsoft.com/en-us/azure/active-directory-domain-services/password-policy

GSuite is very similar

- https://imgur.com/a/ySx60OA 

- google's docs on "secure passwords" https://support.google.com/accounts/answer/9094506?hl=en

Okta mentions a default password policy -- but i couldn't find it online (https://help.okta.com/en/prod/Content/Topics/Security/Security_Policies.htm). 

I think microsoft's policy for windows 10 could be a good starting point.

BMichael00-alt:
Yeah if I could get it back then I'd be able to see the results of my life work on the next three weeks or so to get the best quality of it in my own way and my work is not just about the sexual aspect but the fact that I can do that with the other girls is just as important as a member of the time to get it done with a little bit more than I can do with my own way of doing this for me and u know what I'm trying to say that i'm a good team in my room with a lot on the road to get it done with a smile

[...]

PR #274, opened 2020-06-29
Add password rule for apple.com

jumoru:
### Overall Checklist

#### for password-rules.json

### Description

According to Apple's page for creating a new Apple ID ([link](https://appleid.apple.com/account#!&page=create)):

- at least 8 characters
- not more than 63 characters
- at least one upper case letter, one lower case letter and one digit

Bildschirmfoto 2020-06-29 um 19 37 28
Bildschirmfoto 2020-06-29 um 19 42 18

All ascii-printable characters are allowed as well (determined by testing).

Together this translates into:
`minlength: 8; maxlength: 63; required: lower; required: upper; required: digit; allowed: ascii-printable;`

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -26,6 +26,9 @@
     "angieslist.com": {
         "password-rules": "minlength: 6; maxlength: 15;"
     },
+    "apple.com": {
+        "password-rules": "minlength: 8; maxlength: 63; required: lower; required: upper; required: digit; allowed: ascii-printable;"
+    },
     "artscyclery.com": {
         "password-rules": "minlength: 6; maxlength: 19;"
     },


rmondello:
I’m thinking about this one in the context of the [ongoing conversation](https://github.com/apple/password-manager-resources/issues/134) about what to include in the project and not. These rules seem eminently reasonable to me, and would be satisfied, I suspect, by many password manager’s generated passwords. But maybe not.

But what’s the harm of incorporating a reasonable known good rule? Any password manager using this data can decide not to ship data that they believe describes a policy that is trivially satisfiable. Parsing a rule allows for making that kind of decision. Maybe we want to give up on the notion of this being a shame list, and we’ll document what we can.

+@bradcush @igor-makarov, I’m interesting in your perspectives here.

Very honestly looking for feedback! (Also, I don’t at all care if apple.com makes it onto the list. If it’s helpful for it to be there, it should be there. :-))

igor-makarov:
Here's my take on it - currently, there has been ~130 PRs adding valuable data. Relax the rules a bit, you might be getting more PRs but the data quality won't go up. Plus, the additional PRs would need to get reviewed.

What I think is really needed is a clear definition of "obvious password rules" that don't need documenting.  
My suggestion is generate a short list of good basic password examples (like the ones Safari does) and test the added password rules against those using an automation test. If the new rule doesn't fail any of the basic passwords, it might be too basic to include.

bradcush:
@rmondello Personally I think it makes sense to include rules that go beyond just a simple `minlength` property. If a password manager can use the role to provide a better experience in some way then it's valuable. As an example it's not uncommon to have your default settings for password generation set to something that most sites accept which can force you to compromise a bit. Whether it's simply a defined length limit or common characters that are usually required, these all help to better inform the password manager on what to generate and can translate to better security per site.

jumoru:
@rmondello I can totally see why this rule could seen as something one would expect a password manager to generate. Overall, my motivation to add this rule was driven by my impression that this isn't a "shame" list, but a list that helps making passwords more secure and to increase the user experience of password managers.

Due to this I would side with @bradcush. I think it would be reasonable to document any password rule that has some restrictions other than let's say `minlength <= 8`.

PR #277, opened 2020-07-02
added worldlink.com.np

theshapguy:
### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -290,6 +290,10 @@
         "luxilon.com"
     ],
     [
+        "worldlink.com.np",
+        "nettv.com.np"
+    ]    
+    [
         "wsj.com",
         "dowjones.com"
     ]


quirks/websites-with-shared-credential-backends.json
```diff
@@ -289,6 +289,10 @@
         "evoshield.com",
         "luxilon.com"
     ],
+    [
+        "worldlink.com.np",
+        "nettv.com.np"
+    ]
```
rmondello, line 295: Can you provide some information about how you determined that these websites are owned by the same entity and can be logged into with the same credentials? Can you provide URLs of the login pages on each of these websites?

theshapguy: So, I'm WorldLink (ISP) and NetTV (TV Subscription) share the same credentials. You can only your NetTV subscription through Worldlink and have to login with your Worldlink credentials. 

You can view the link in the banner picture at https://www.worldlink.com.np/iptv.php. "Watch it on NetTV".

rmondello: This helps a lot! Thanks!

PR #278, opened 2020-07-02
Added www4.irs.gov password rules

mleibner:
This **www4.irs.gov** entry covers a single IRS.gov login system that spans a few subdomains like **sa.www4.irs.gov** & **la.www4.irs.gov**.

_For example, the service linked to from https://www.irs.gov/account lives under "**sa.www4.irs.gov**", and the services linked to from https://www.irs.gov/e-services live under "**la.www4.irs.gov**", and these both share the same login system + password rules._

There are other irs.gov subdomains - outside of www4 - which have different login systems and password requirements; those are not covered in this pull request.

This request is being submitted since the system only allows the **specific symbols** listed in the password rules. Attempts to use other characters will result in failure.

___

**Screenshot of www4.irs.gov password rules:**
www4 irs gov pw rule screenshot
___
**Plaintext reading of www4.irs.gov password rules:**
> Password Rules:
• Between 8 and 32 characters long.
• Must contain at least one numeric and one special character (!@#$%&*).
• At least one uppercase and at least one lowercase letter.
___
**Example showing that other symbols (^) not in their list do not fulfill the "special character" test:**
special char fail

___

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -345,6 +345,9 @@
     "wsj.com": {
         "password-rules": "maxlength: 15;"
     },
+    "www4.irs.gov": {
+        "password-rules": "required: lower; required: upper; required: digit; required: [!#$%&*@]; minlength: 8; maxlength: 32;"
+    },
     "xfinity.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower, upper; required: digit;"
     },
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -346,7 +346,7 @@
         "password-rules": "maxlength: 15;"
     },
     "www4.irs.gov": {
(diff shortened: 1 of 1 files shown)


rmondello:
Do you think the `exact-domain-match-only` key could be useful here?

rmondello:
(Will merge with an answer to that question.)

mleibner:
I don't think `exact-domain-match-only` should apply since this is actually meant to cover **sa**.www4.irs.gov & **la**.www4.irs.gov (there's nothing deployed at just www4.irs.gov).

My intention was to cover *.www4.irs.gov with these rules,  and specifically **exclude** other subdomains (like https://rpr.irs.gov/ ) which have different password rules.

Hope that makes sense. Thanks!

rmondello:
@mleibner That’s perfect. Thanks for explaining your intention! You’ve done it exactly right.

rmondello:
(Closing the PR was an mis-click.)

PR #290, opened 2020-07-13
Add Discord to shared backends

plttn:
Add discord.com/discordapp.com to shared-credential-backends

While not technically falling into the exact specifics of the requirements to be added to this quirk, it seems to meet enough of them that it would be logical to add.

Discord used to previously use https://discordapp.com as their primary branding URL (having owned https://discord.com for some time), but at the beginning of May switched to redirecting https://discordapp.com to https://discord.com. Anyone with the old URL saved in their password manager would need to adjust it to https://discord.com.

image
image

### Overall Checklist

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -101,6 +101,10 @@
         "dinersclubus.com"
     ],
     [
+        "discord.com",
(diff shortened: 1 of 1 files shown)


rmondello:
Thoughtful change. Thanks @plttn!

PR #291, opened 2020-07-16
Add ali(baba|express) to shared backends

alistra:
### Overall Checklist

[Source](https://www.easyship.com/blog/alibaba-vs-aliexpress)

> Are Alibaba and AliExpress the Same?

> To put it in easy-to-understand terms, Alibaba and AliExpress are like a pair of siblings. They share the same parents, but have their own personalities and identities. While Alibaba and AliExpress are owned by the same company, each has a different purpose, operates differently, and is designed for a different audience. Specifically, Alibaba operates on a B2B model while AliExpress works on a B2C model.

The both pages share a credential backend and they appear as a warning on my iOS 14, so I wanted to address that :)

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -9,6 +9,10 @@
         "banneraetna.myplanportal.com"
     ],
     [
+        "alibaba.com",
+        "aliexpress.com"
+    ],
+    [
         "alltrails.com",
         "alltrails.io"
     ],


rmondello:
Are there URLs on both domains where I can log in with the same credentials?

alistra:
https://passport.alibaba.com/icbu_login.htm is the login page on one
the other just has a modal window on https://www.aliexpress.com

Not sure if I should provide the specific subdomain, it is probably better to keep it general, so it is less volatile.

rmondello:
I created an aliexpress account, and I could not use it to sign into alibaba. I’m not sure if this is a valid PR.

alistra:
Whenever I change the password on any of the sites, the other one changes too. Maybe it is some kind of linked account, but you need initial activation? Seems weird anyway

bradcush:
@rmondello I would agree with you here. I've done the opposite flow and created an account on https://passport.alibaba.com/icbu_login.htm and was unable to login to https://www.aliexpress.com. These sites also have separate SSL certificates stating different DNS entries.

rmondello:
Closing this for now. Please reopen, any time, if new information comes to light!

Ramias1:
My aliexpress account may not work across "normal" alibaba but passport.alibaba.com and webatm.alibaba.com are the same as my Aliexpress password.

PR #294, opened 2020-07-19
Add password rules for ventrachicago.com

davidjohnson91:
### Overall Checklist

#### for password-rules.json

When changing a password, the rules are outlined as:
> Password must have a minimum of 8 characters and include one upper case letter, one lower case letter, and at least one number or one of these characters (!@#$%^) No other characters are allowed.

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -122,6 +122,7 @@
     "udel.edu": "https://udapps.nss.udel.edu/myUDsettings/password",
     "umsystem.edu": "https://password.umsystem.edu/reset/",
     "usaa.com": "https://www.usaa.com/inet/ent_auth_password/pages/ChangePasswordPage",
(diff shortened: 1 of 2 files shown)


jumoru:
Looks fine to me. The only little thing I've noticed is that, given they require a number **or** (!@#$%^), the password rule could be made a bit less restrictive (`required: digit, [!@#$%^];` instead of  `required: digit; required: [!@#$%^];`).

rmondello:
@davidjohnson91 What do you think about making this?

```
required: lower; required: upper; required: digit, [!@#$%^];
```

As @jumoru suggests.

davidjohnson91:
Thanks for the feedback everyone, I'm still learning the syntax for these rules a bit. I also verified that the description is correct and that `digit || [!@#$%^]` causes a password to be accepted.

rmondello:
@davidjohnson91 Learning is good. :)

Thanks for the PR!

PR #295, opened 2020-07-20
Add password rules for walkhighlands.co.uk

mlabuhn:
Added a rule for https://www.walkhighlands.co.uk, using the requirements stated on their own account settings page (https://www.walkhighlands.co.uk/Forum/ucp.php?i=profile&mode=reg_details):

"Password must be between 9 and 15 characters long, must contain letters in mixed case and must contain numbers."

I have established through experimentation that beyond the minimum complexity requirements, the site also accepts all characters in the 'special' character class. I have not tested the full 'ascii-printable' or 'unicode' character sets, these may or may not be fully supported.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -367,7 +367,7 @@
         "password-rules": "maxlength: 6; max-consecutive: 3; allowed: digit;"
     },
     "walkhighlands.co.uk": {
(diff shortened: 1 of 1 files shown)


mlabuhn:
Submitting above change for review

rmondello:
Thanks @mlabuhn!

PR #308, opened 2020-08-13
Move hyphens to front of character class

Cldfire:
According to [the spec](https://github.com/whatwg/html/issues/3518), hyphens are only valid at the very beginning of a character class:

```
The dash character (-) is reserved as a special character. To list '-' as a literal character it must appear immediately after the opening square bracket '['.
```

Our parser over at @1Password caught a couple rules in the list that have the `-` in non-valid places, so I moved them to the start of the character class in this PR.

### Overall Checklist


rmondello:
Thanks @Cldfire! I’m still working on getting us some automation that’ll check this. (I know it’s been a while. The problems are non-technical.)

PR #331, opened 2020-08-25
walmart.com: remove min/max length attributes, add allowed characters to exclude not-allowed space

sashkab:
### Overall Checklist

#### for password-rules.json

Password change form at https://www.walmart.com/account/profile:

screenshot_2020 08 25_184727

Account registration page:

screenshot_2020 08 25_185211

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -415,7 +415,7 @@
         "password-rules": "minlength: 9; maxlength: 15; required: lower; required: upper; required: digit; allowed: special;"
     },
     "walmart.com": {
-        "password-rules": "minlength: 6; maxlength: 12;"
+        "password-rules": "allowed: lower, upper, digit, [-(~!@#$%^&*_+=`|(){}[:;\"'<>,.?]];"
     },
     "web.de": {
         "password-rules": "minlength: 8; maxlength: 40; allowed: lower, upper, digit, [-<=>~!|()@#{}$%,.?^'&*_+`:;\"[]];"


rmondello:
The fact that they’re declaring a minlength and maxlength on the <input> means that we might not need this rule at all in the repo. What does everyone think? @bradcush

Scetch:
> The fact that they’re declaring a minlength and maxlength on the means that we might not need this rule at all in the repo. What does everyone think? @bradcush

I agree with this one, I'm not sure this needs to be included in the repo either. Over at @1Password we look at `minlength` and `maxlength` on fields in addition to the `passwordRules` to determine if we need to modify our default password rules.

sashkab:
Looks like they don't accept space as valid character.

screenshot_2020 08 25_212813

Does it mean, we need to use something like this?

```
allowed: lower; allowed: upper; allowed: digit; allowed: [(-~!@#$%^&*_+=`|(){}[:;"'<>,.?]];
```

crypt096 (approved):

rudyrichter:
I think @sashkab is right, we don't need min/max, but the "pattern" described in the html doesn't line up with [https://developer.apple.com/password-rules/](`passwordRules`)

rmondello:
Thanks everyone for the help! I think this website is worthy of a passwordRule:

```
allowed: lower; allowed: upper; allowed: digit; allowed: [-(~!@#$%^&*_+=`|(){}[:;"'<>,.?]];
```

(Note that compared to @sashkab’s posted rule above, I move the `-` to be the first character of the character class.)

Why this rule:
- The website is giving us enough information about length.
- The pattern attribute is great for a validating user-typed text in a field, it’s cumbersome for password managers to work out how to meet. passwordRules were designed to be used by password managers.

bradcush:
@rmondello I'm in agreement on adding this rule based on your reasoning without the min/max length attributes.

sashkab:
removed min/max length, squashed into a single commit.

rmondello:
Thanks @sashkab!

PR #332, opened 2020-08-27
Add speedway.com

GeoJunkie:
### Overall Checklist

#### for password-rules.json

Tested at https://login.speedway.com/register

Screenshot on 2020-08-27 at 06-38-16


bradcush:
I've verified that the password change URL given redirects to the login page when not logged in and to the appropriate change view when logged in. Regarding the rules I've verified that indeed only digits are accepted and the form actually restricts the user to enter anything other than a digit as a character and any more characters than the max of 8. (It doesn't implement the standard min/max but does use the standard `pattern`)

rmondello:
@bradcush You can merge this if you’d like!

PR #341, opened 2020-09-08
Add password rule for store.nvidia.com

Cldfire:
Adds a rule for https://store.nvidia.com.

This site is somewhat complicated to get to. You have to:

* Go to https://www.nvidia.com/en-us/shop/geforce
* Find an NVIDIA founder's edition card that's in-stock and choose "Add to cart"
* Go to the cart at the top right and choose "Checkout"
* You'll be redirected to a https://store.nvidia.com page
* Choose "Create account"
* The next page will have the password fields on it. You'll need to input fake data for the rest of the shipping form in order to see errors about the password, but you do not need to fill out the credit card number form

Screenshot of the rules:

Screen Shot 2020-09-07 at 8 59 17 PM

It doesn't seem like there's a nice way to express the "3 of 4" requirement here in password rules, so I just went ahead and required one character from all four classes.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -377,6 +377,9 @@
     "splunk.com": {
         "password-rules": "minlength: 8; maxlength: 64; required: lower; required: upper; required: digit; required: [-!@#$%&*_+=<>];"
     },
+    "store.nvidia.com": {
+        "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [-!@#$%^*~:;&><[{}|_+=?]];"
+    },
     "sulamericaseguros.com.br": {
         "password-rules": "minlength: 6; maxlength: 6;"
     },


quirks/password-rules.json
```diff
@@ -377,6 +377,9 @@
     "splunk.com": {
         "password-rules": "minlength: 8; maxlength: 64; required: lower; required: upper; required: digit; required: [-!@#$%&*_+=<>];"
     },
+    "store.nvidia.com": {
```
Cldfire, line 380: Also, quick clarification here. Our README states:

```
When a domain is listed in quirks/password-rules.json, it means that that domain and all of its subdomains use the rule.
```

I'm assuming this means that when specifying `store.nvidia.com` the rule is allowed to match `*.store.nvidia.com`, as opposed to `*.nvidia.com`, and if true I'll open a PR to make that explicitly clear in the README.

rmondello: That’s exactly right, @Cldfire!

rmondello (approved):

PR #363, opened 2020-09-29
Add cigna.co.uk to password rules

Flameeyes:
This is different from the (US?) cigna.com website rules because no symbols are allowed at all.

See screenshot:

image

Namely no special characters are allowed.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -92,6 +92,9 @@
     "chase.com": {
         "password-rules": "minlength: 8; maxlength: 32; required: lower, upper; required: digit; required: [!#$%+/=@~]; max-consecutive: 2;"
     },
+    "cigna.co.uk": {
+        "password-rules": "minlength: 8; maxlength: 12; required: lower; required: upper; required: digit;"
+    },
     "cigna.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -92,6 +92,9 @@
     "chase.com": {
         "password-rules": "minlength: 8; maxlength: 32; required: lower, upper; required: digit; required: [!#$%+/=@~]; max-consecutive: 2;"
     },
+    "cigna.co.uk": {
+        "password-rules": "minlength: 8; maxlength: 12; required: lower, upper, digit;"
```
Cldfire, line 96: The screenshot makes it sound like their password rules require at least one number, at least one lowercase letter, and at least one uppercase letter, all at the same time. That requirement would need to be expressed with:

```
required: lower; required: upper; required: digit;
```

This ensures that at least one of each character type will be present (whereas `required: lower, upper, digit;` does not make that guarantee).

Flameeyes: Thanks, I didn't realize that difference. Updated now.

Cldfire (approved):
Perfect, thank you very much for the contribution! 😄

PR #365, opened 2020-10-13
Add foursquare/swarm to shared-credential-backends

fbartho:
### Overall Checklist

I'm a longtime user of foursquare.com ["Superuser Level 2"]((https://foursquare.com/fbartho), and Swarmapp.com is gameified checkin application was split out from the main foursquare product but otherwise shares the same account.

I was updating my passwords and I noticed the following warning in Safari on macOS that clearly indicates that Safari doesn't know these are related domains.

Screenshot of Safari password manager interface with alerts that my credentials are duplicated across these two domains

There is only 1 account system for these two domains, but the Swarmapp.com website does not have a unique password management url that I was able to discover. Instead they recommend you interact from within the app, but for the purposes of this PR I've configured the URL to point at the main foursquare password-management page -- that said, I don't think Safari will automatically catch when the users change passwords in this way so I'm open to suggested changes.


Cldfire (approved):
The additions here look great, thanks!

I'll let someone else comment on the Safari-specific questions.

rmondello (approved):

rmondello:
I’ll have to check on Safari’s “update password” behavior for associated domains like this. If it didn’t do the right thing, it certainly could.

Thanks for the PR @fbartho!

PR #366, opened 2020-10-13
[Feature] JSON Formatting Tool

fbartho:
### Overall Checklist

## Context

While working on #365 I ran into some dev-facing annoyances while working with the JSON files

- Half the files had trailing newlines, and half did not
- One of the files had some junk whitespace that had been accidentally contributed
- My editor autodetected the spacing rules in one file, so that was great, but for another file it tried to reformat about half the entries to be in a denser "on-one-line" format.

### Changes

- [x] Add a ruby-based formatting tool that matches the syntax / decisions in existing ruby scripts available in this repo
- [x] Fix noticed JSON Formatting Issues
- [x] Add an editorconfig.org config file to help the cluster of editors that automatically respect such things


rmondello:
Hey @fbartho! Thank you for this. This looks almost entirely perfect for merging as-is to me, but two questions:

1. How would you feel about two space indentation for the ruby file? That would feel more idiomatic to me, for ruby.
2. Genuine question, because I don’t have a lot of experience with these: Do you think the .editorconfig should go in the root of the project? Is that typically what projects do? Or should we create some subdirectory for it?

fbartho:
@rmondello 

1. Re: spacing in Ruby: I thought I matched the spacing rules demonstrated in the other Ruby files that existed in this repo before me! I'm a tab person, myself, but as long as they're all consistent, I can survive! (Let's just add an entry in editorconfig for Ruby files with whatever the decision is)
2. Re: editorconfig location: I'm not an expert at it, but I believe you must have one at the root. You can also have others for sub￼trees of folders, but I believe it's not designed to be located at an arbitrary path.

Cldfire:
>     2. I'm not an expert at it, but I believe you must have one at the root.

Yep, this is my understanding as well. https://editorconfig.org/ says:

> When opening a file, EditorConfig plugins look for a file named .editorconfig in the directory of the opened file and in every parent directory. A search for .editorconfig files will stop if the root filepath is reached or an EditorConfig file with root=true is found.

rmondello (approved):

rmondello:
@fbartho 

1. You’re totally right. You’re just matching the style that’s already in the project. Rather than hold up your improvement on bikeshedding about spacing, I merged your change. :)

2. Sounds good. And thanks @Cldfire!

PR #367, opened 2020-10-13
adding password rules for verizonwireless.com

axayjha:
### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -444,7 +444,7 @@
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit, [!@#$%^];"
     },
     "verizonwireless.com": {
-        "password-rules": "minlength: 8; maxlength: 20; required: digit; required: lower, upper;"
+        "password-rules": "minlength: 8; maxlength: 20; required: digit; required: lower, upper; allowed: unicode;"
     },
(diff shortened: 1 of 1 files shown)


Cldfire:
I can't tell from the screenshot or their documentation, and I can't seem to find a way to get to the password part of the signup page; do you know if they allow any kind of special characters? If so, it would be great to allow them in the password rule as well.

axayjha:
You can get to password part of of the signup page via here (they will send registration link to your email): https://login.verizonwireless.com/accessmanager/public/c/reg/guest/start?clientId=telematics.

And yes they allow special characters; in fact, they allow all unicode characters.

Cldfire:
> You can get to password part of of the signup page via here (they will send registration link to your email): https://login.verizonwireless.com/accessmanager/public/c/reg/guest/start?clientId=telematics.
> 
> And yes they allow special characters; in fact, they allow all unicode characters.

Oh perfect, thanks for that link!

It does seem to allow all special / unicode characters. Let's add `allowed: unicode;` to the rule to express that.

axayjha:
Sure, will do.

axayjha:
Added

Cldfire (approved):

Cldfire:
Awesome. Thank you!

PR #369, opened 2020-10-14
Add Wii-Homebrew PasswordChange URL

InternalLoss:
### Overall Checklist

Wii-Homebrew Board's change password URL is from the User drop-down -> Account Management; the user will see an "You are not authorised to view this page" if not logged in, which is a concern, but this still makes it easier to access the password change page rather than having to find the relevant page in their settings, especially when it's not named something simple as "Password".


rmondello (approved):

InternalLoss:
Thanks for the quick merge! I asked previously on the other issue, but is it public how quickly changes are added into iOS/macOS (i.e. is it just bundled into new app updates/sent some other way)? If it's not that's no worry, just curious ^^

PR #373, opened 2020-10-29
Add google.com - Password Rule

gobinathm:
Follow Up of #36  

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -218,6 +218,9 @@
     "gmx.net": {
         "password-rules": "minlength: 8; maxlength: 40; allowed: lower, upper, digit, [-<=>~!|()@#{}$%,.?^'&*_+`:;\"[]];"
     },
+    "google.com": {
+        "password-rules": "minlength: 8; allowed: lower, upper, digit, [-!\"#$%&'()*+,./:;<=>?@[^_{|}~]];"
+    },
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -218,6 +218,9 @@
     "gmx.net": {
         "password-rules": "minlength: 8; maxlength: 40; allowed: lower, upper, digit, [-<=>~!|()@#{}$%,.?^'&*_+`:;\"[]];"
     },
+    "google.com": {
+        "password-rules": "minlength: 8; allowed: lower, upper, digit, [-!\"#$%&'()*+,./:;<=>?@[^_{|}~]];"
```
rmondello, line 222: Are you attempting to call out any ascii-printable character here that Google does not allow?

rmondello (changes requested):

quirks/password-rules.json
rmondello: I’m guessing space?

rmondello (changes requested):

quirks/password-rules.json
gobinathm: @rmondello correct spaces are not allowed hence the specific set of special characters are included.

rmondello (approved):

PR #375, opened 2020-11-02
Update list of websites with shared credentials backend

MartinDevi:
### Overall Checklist

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -314,10 +314,6 @@
         "disneyplus.com"
     ],
     [
-        "sodexo.com",
-        "sodexopass.fr"
-    ],
-    [
         "slcl.overdrive.com",
         "slcl.org"
     ],
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -293,10 +293,6 @@
         "getpocket.com"
     ],
     [
-        "pornhub.com",
-        "pornhubpremium.com"
-    ],
-    [
         "probikeshop.fr",
         "bikeshop.es",
         "probikeshop.it",
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -5,8 +5,67 @@
         "americanairlines.jp"
     ],
     [
+        "airfrance.com",
+        "flyingblue.com"
+    ],
+    [
         "aetna.com",
         "banneraetna.myplanportal.com"
+    ],
+    [
+        "airbnb.com.ar",
+        "airbnb.com.au",
+        "airbnb.at",
+        "airbnb.be",
+        "airbnb.com.bz",
+        "airbnb.com.bo",
+        "airbnb.com.br",
+        "airbnb.ca",
+        "airbnb.cl",
+        "airbnb.com.co",
+        "airbnb.co.cr",
+        "airbnb.cz",
+        "airbnb.dk",
+        "airbnb.com.ec",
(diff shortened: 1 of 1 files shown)


quirks/websites-with-shared-credential-backends.json
```diff
@@ -17,6 +76,7 @@
         "amazon.com.au",
         "amazon.com.br",
         "amazon.ca",
+        "amazon.cn",
```
rmondello, line 79: Are we sure that amazon.cn shares a credential backend with the rest?

rmondello (changes requested):

quirks/websites-with-shared-credential-backends.json
MartinDevi: Huh, I actually can't log in on that website with my amazon.com credentials, so no. It does seem to be owned by Amazon, but I guess they didn't merge their database of user accounts.

My bad, I'll remove it. Thanks, good catch.

MartinDevi: amazon.co.jp was mistaken as well it seems

quirks/websites-with-shared-credential-backends.json
```diff
@@ -363,5 +438,9 @@
     [
         "wsj.com",
         "dowjones.com"
+    ],
+    [
+        "yahoo.com",
```
Cldfire, line 443: This one seems complicated. https://techcrunch.com/2019/03/04/after-more-than-10-years-flickr-frees-its-login-system-from-yahoo/ seems to indicate that going forward this transform will no longer be accurate.

MartinDevi: You're right, I just tested, I can totally create a Flickr account which doesn't work to log in on Yahoo.

MartinDevi: Removed

Cldfire (approved):
Everything else looks reasonable to me 🙂 I checked on the airbnb.com <=> airbnb.jp mapping and that one holds true, so I'm assuming the rest of them will work as well.

I'll leave it for @rmondello to take another look 😄

quirks/websites-with-shared-credential-backends.json
```diff
@@ -250,6 +313,10 @@
         "go.com",
         "disneyplus.com"
     ],
+    [
+        "sodexo.com",
```
rmondello, line 317: Where can I log in on www.sodexo.com?

MartinDevi: Couldn't find a login form either... removed it then. Sorry about that.

bradcush:
For the current domains in the listed after some changes were made of done a few manual checks.
- "airfrance.com" and "flyingblue.com" were tested and confirmed to use the same login information (The SSL certificates are different as the sites seem to be associated and partnered with each other. There are also potentially other carrier logins they support mentioned at the bottom of the page on https://www.flyingblue.us/en/home/explorer we should consider checking and adding in the future.)
- I didn't try all the airbnb sites in the PR but everyone I did (around 5) allowed the same login which was contained on the relevant domain itself. I've also checked around 10 of the site certificates and they are all using 1 of 2 shared certificates, either for www.airbnb.com or www.airbnb.at.
- "fnac.com" and "fnacspectacles.com" were tested manually and allow the same login. (Both are using their own unique SSL certificate specific to their domains)
- I was unable to test "wellsfargo.com" and "wellsfargoadvisors.com" but I found while the two sites have their own unique SSL certificates they do link to each other. (If someone else can confirm them in terms of login that would be great)

bradcush (approved):

rmondello (approved):

PR #377, opened 2020-11-11
Add password rule for greatwestlife.com

Cldfire:
Adds a password rule for https://gwl.greatwestlife.com/MyRegistration?lang=en_US

I scoped the rule to the `gwl` subdomain; as far as I can tell that seems to be the best thing to do, although it's difficult to find out for sure.

Screenshot of the rules on the site:

Screen Shot 2020-11-11 at 12 04 57 PM

I omitted `maxlength` from the rule because the field on the site has the attribute on it already.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -239,6 +239,9 @@
     "google.com": {
         "password-rules": "minlength: 8; allowed: lower, upper, digit, [-!\"#$%&'()*+,./:;<=>?@[^_{|}~]];"
     },
+    "gwl.greatwestlife.com": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [-!#$%_=+<>];"
+    },
     "guardiananytime.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -239,6 +239,9 @@
     "google.com": {
         "password-rules": "minlength: 8; allowed: lower, upper, digit, [-!\"#$%&'()*+,./:;<=>?@[^_{|}~]];"
     },
+    "gwl.greatwestlife.com": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%-_=+<>];"
```
rmondello, line 243: `-` needs to be the first character of the class.

rmondello (changes requested):

quirks/password-rules.json
Cldfire: Good catch!

rmondello (approved):

PR #380, opened 2020-11-18
Add Wayfair family of websites with shared credentials

niorio:
### Overall Checklist

Noticed that safari was flagging Wayfair and Allmodern as reused passwords. I did some research and found that there's a whole family of Wayfair sites that use the same credential system. All these domains have www.wayfair.com SSL certs.


Cldfire (approved):
This looks great, thank you! There's definitely ample evidence that the sites are interconnected.

rmondello:
Thanks @niorio!

PR #383, opened 2020-11-22
add comdirect.de password rule

Michi01:
add comdirect.de password rule

### Overall Checklist
- [ X] I agree to the project's [Developer Certificate of Origin](https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md)
- [X ] The top-level JSON objects are sorted alphabetically
- [ X] There are no [open pull requests](https://github.com/apple/password-manager-resources/pulls) for the same update

#### for password-rules.json
- [ X] The given rule isn't particularly standard and obvious for password managers
- [X ] Generated passwords have been tested from this rule using the [Password Rules Validation Tool](https://developer.apple.com/password-rules/)
- [X ] Information has been included about the website's requirements (eg. screenshots, error messages, steps during experimentation, etc.)
- [X ] The PR isn't documenting something that would be a common practice among password managers (e.g. minimal length of 6)
Screen Shot 2020-11-22 at 14 40 27

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -125,6 +125,9 @@
     "comcastpaymentcenter.com": {
         "password-rules": "minlength: 8; maxlength: 20; required: lower, upper; required: digit; max-consecutive: 2;"
     },
+    "comdirect.de": {
+         "password-rules": "minlength: 6; maxlength: 6; required: digit;"
+    },
     "comed.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; allowed: [-~!@#$%^&*_+=`|(){}[:;\"'<>,.?/\\]];"
     },


Cldfire:
Hmm, it's hard for me to tell if this is an accurate rule for the site or not. Based on a translation of the above screenshot to English, it appears to be correct (the text there is stating that the new pin must be six numeric digits). However, other pages on the site state different requirements.

The login page at https://kunde.comdirect.de/lp/wt/login, in German:

Screen Shot 2020-11-24 at 9 50 23 AM

In English:

Screen Shot 2020-11-24 at 9 50 13 AM

This appears to be differentiating between the PIN pictured in the original screenshot and a different type of password that can contain letters.

Also, the registration page at https://www.comdirect.de/pbl/https/member/ssomanager/personalization/registration/StartRegistrationRH.do, in German:

Screen Shot 2020-11-24 at 9 57 00 AM

In English:

Screen Shot 2020-11-24 at 9 56 53 AM

Here we see very different password requirements; a password generated according to the rule being added in this PR wouldn't even work here.

It unfortunately seems like this site has more than one rule for its passwords under the same domain?

Michi01:
My PR is for the banking of Comdirect, you Second link seem to be for the User Community
I get redirected to  https://nutzer.comdirect.de/pbl/https/member/ssomanager/personalization/registration/StartRegistrationRH.do If I follow your URL. I was not aware, that you can use the community without having a Bank Account there. But I also don’t find any informations how to handle such cases here?

rmondello:
So there’s both a password and a PIN for the website, with different rules?

Michi01:
Yes it looks like this... online banking, which is the main purpose of the website, only has the PIN. The user Community passwords may also contain other characters.

rmondello:
Are there differences in sub domain we can use to differentiate which part of the website is PIN-based and which is password-based?

Michi01:
I‘ve checked it, no different Subdomains

rmondello:
In case that, it seems more harmful than helpful to add this rule.

PR #304, opened 2020-08-05
Add a rule for benjerry.com

rmondello:
### Overall Checklist
- [X] I agree to the project's [Developer Certificate of Origin](https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md)
- [X] The top-level JSON objects are sorted alphabetically
- [X] There are no [open pull requests](https://github.com/apple/password-manager-resources/pulls) for the same update

#### for password-rules.json
- [X] The given rule isn't particularly standard and obvious for password managers
- [X] Generated passwords have been tested from this rule using the [Password Rules Validation Tool](https://developer.apple.com/password-rules/)
- [X] Information has been included about the website's requirements (eg. screenshots, error messages, steps during experimentation, etc.)
- [X] The PR isn't documenting something that would be a common practice among password managers (e.g. minimal length of 6)

Screen Shot 2020-08-02 at 2 24 51 PM

This is one of the more elaborate sets of rules I’ve seen. Conforming to them is very frustrating. I verified by hand that the website accepts all special characters.

PR #395, opened 2020-12-11
Improved consistency of password-rules.json

CasualDeveloper:
A. Rearranged rules (when present) into the following order:
1. minlength
2. maxlength
3. max-consecutive
4. required
5. allowed

B. Rearranged "required: " and "allowed: " rule records in the following order:
lower > upper > digit > special.

C. Removed extraneous "allowed: " rules if they were already included as a "required: " rule.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -66,7 +66,7 @@
         "password-rules": "minlength: 8; maxlength: 16; required: lower, upper; allowed: digit, special;"
     },
     "benjerry.com": {
-        "password-rules": "required: upper; required: digit; required: special; allowed: lower;"
+        "password-rules": "required: upper; required: upper; required: digit; required: digit; required: special; required: special; allowed: lower;"
     },
     "bestbuy.com": {
         "password-rules": "minlength: 20; required: lower; required: upper; required: digit; required: special;"
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -60,7 +60,7 @@
         "password-rules": "minlength: 8; maxlength: 8; required: lower; required: upper; required: digit;"
     },
     "bankofamerica.com": {
-        "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 3; required: upper; required: lower; required: digit; allowed: [-@#*()+={}/?~;,._];"
+        "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 3; required: lower; required: upper; required: digit; allowed: [-@#*()+={}/?~;,._];"
     },
     "battle.net": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower, upper; allowed: digit, special;"
@@ -393,7 +393,7 @@
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [\"!@#$%^&*(){}[]];"
     },
     "prepaid.bankofamerica.com": {
-        "password-rules": "minlength: 8; maxlength: 16; required: upper; required: lower; required: digit; required: [!@#$%^&*()+~{}'\";:<>?];"
+        "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [!@#$%^&*()+~{}'\";:<>?];"
     },
     "propelfuels.com": {
         "password-rules": "minlength: 6; maxlength: 16;"
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -147,7 +147,7 @@
         "password-rules": "minlength: 7; maxlength: 18; required: lower, upper; required: digit;"
     },
     "cvs.com": {
-        "password-rules": "minlength: 8; maxlength: 25; required: lower, upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@#$%^&*()];"
+        "password-rules": "minlength: 8; maxlength: 25; required: lower, upper; required: digit; allowed: [!@#$%^&*()];"
     },
     "dailymail.co.uk": {
         "password-rules": "minlength: 5; maxlength: 15;"
@@ -258,7 +258,7 @@
         "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit;"
     },
     "hotels.com": {
-        "password-rules": "minlength: 6; maxlength: 20; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789@$!#()&^*%];"
+        "password-rules": "minlength: 6; maxlength: 20; required: digit; allowed: lower, upper, [@$!#()&^*%];"
     },
     "hotwire.com": {
         "password-rules": "minlength: 6; maxlength: 30; allowed: lower, upper, digit, [-~!@#$%^&*_+=`|(){}[:;\"'<>,.?]];"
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -60,16 +60,16 @@
...
-        "password-rules": "minlength: 8; maxlength: 20; required: upper; required: lower; required: digit; max-consecutive: 3; allowed: [-@#*()+={}/?~;,._];"
+        "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 3; required: lower; required: upper; required: digit; allowed: [-@#*()+={}/?~;,._];"
     },
     "battle.net": {
-        "password-rules": "required: upper,lower; allowed: digit,special; minlength: 8; maxlength: 16;"
+        "password-rules": "minlength: 8; maxlength: 16; required: lower, upper; allowed: digit, special;"
     },
     "benjerry.com": {
-        "password-rules": "required: upper; required: upper; required: digit; required: digit; required: special; required: special; allowed: lower;"
+        "password-rules": "required: upper; required: digit; required: special; allowed: lower;"
```
Cldfire, line 69: This isn't semantically equivalent to the rule as it was previously.

The previous rule was:

```
required: upper; required: upper; required: digit; required: digit; required: special; required: special; allowed: lower;
```

This requires the password to have:

* At least two uppercase letters
* At least two digits
* At least two special characters

However, the rule as you've written it is more relaxed, requiring:

* At least one uppercase letter
* At least one digit
* At least one special character

If you look at the PR that added this rule (#304) you'll notice that the relaxed requirements wouldn't be appropriate for this site.

Cldfire (changes requested):
The consistent formatting is really nice, thank you!

I'll leave this for others to review and comment on as well.

rmondello:
I think we should codify what our preferred order is, and then have me update the Validation Tool to, by default, output rules in that order. (So it might take us a little bit to get this merged, but it'll be worth it!)

quirks/password-rules.json
CasualDeveloper: 🤦🏻‍♂️ Oops, looks like I misunderstood the following line from the [Apple Developer document](https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules) then: 
> You can combine these keywords to form your rule. _Duplicate property values_, empty character classes, and properties without a value _are ignored_.

Thanks for catching that, I'll fix it.

CasualDeveloper:
> If you look at the PR that added this rule (#304) you'll notice that the relaxed requirements wouldn't be appropriate for this site.
Fixed, thank you! This has been a learning experience for me 👍🏻 


> I think we should codify what our preferred order is, and then have me update the Validation Tool to, by default, output rules in that order. (So it might take us a little bit to get this merged, but it'll be worth it!)
Sounds good! I wouldn't mind changing the order too, if anyone has any ideas. The order I proposed was just the way my brain worked, lol 😆

quirks/password-rules.json
Cldfire: Yeah, I agree it can be confusing; the `required` property is the exception to that.

rmondello:
My slight preference is for:

minlength
maxlength
required
allowed
max-consecutive

Because that puts the less-commonly-used item, `max-consecutive`, at the end of lines. Thoughts?

Cldfire:
My *slight* preference would be:

minlength
maxlength
max-consecutive
required
allowed

Reasoning being, `required` and `allowed` are the two properties you're going to spend the most time referencing when reading a rule (as they take the most effort to understand), so it's nice to have the guarantee that they'll be anchored to the rightmost portion of the rule.

Your reasoning makes sense as well, though. I'm interested to hear what others think!

CasualDeveloper:
Both of you make good points!

Personally, however, my thought process groups `minlength`, `maxlength`, and `max-consecutive` together because they all use a single integer as their value.

To me, the equivalence between key name and value makes them simpler to parse as a human; while `required:` and `allowed:` can have arrays as their value, which requires the brain to spend a little more processing time (however slight).

rmondello:
I’m okay with this.

I’ll see if I can whip up a change to the developer.apple.com tool.

rmondello:
Okay, pushed a change to https://developer.apple.com/password-rules/ to make the order match the proposed order here.

Feel free to:
1. Continue getting this PR reviewed.
2. Test https://developer.apple.com/password-rules/ and let me know if you run into any problems.

CasualDeveloper:
Hi Ricky, I'm not seeing the changes being live on [https://developer.apple.com/password-rules/](https://developer.apple.com/password-rules/) yet:
Screenshot 2020-12-23 at 10 07 45 PM

rmondello:
@CasualDeveloper I’m not sure how you’d be able to tell with those rules.
Screen Shot 2020-12-23 at 9 37 58 AM

Cldfire:
It sounds like @CasualDeveloper was referring to sub-sorting of the `required` property mentioned in the PR description:

```
B. Rearranged "required: " and "allowed: " rule records in the following order:
lower > upper > digit > special.
```

rmondello:
> It sounds like @CasualDeveloper was referring to sub-sorting of the `required` property mentioned in the PR description:
> 
> ```
> B. Rearranged "required: " and "allowed: " rule records in the following order:
> lower > upper > digit > special.
> ```

Oh gosh! I haven’t implemented that yet! But I don’t think we have to block this change on me doing that. Let’s continue with reviewing this and get it in. :-)

Cldfire:
I threw https://github.com/Cldfire/password-rules-checker together, and according to it all the rules as they are in this PR:

* Are valid (they parse successfully)
* Are semantically equivalent to the rules as they were previously

So everything looks good to me :). (Feel free to run the tool / play around with it to verify the results.)

It did make a couple suggestions which I'll echo here.

Cldfire:
Oh, github won't let me comment on lines that haven't been modified in this PR ☹️. Anyway, the tool is saying the following:

```
sfwater.org: the `allowed` property for this rule can be shortened to: [Upper, Lower, Custom(['!', '#', '$', '%', '(', ')', '*', '+', '.', ':', ';', '?', '@', '^', '_', '{', '}'])]
marriott.com: the `allowed` property for this rule can be shortened to: [Custom(['!', '#', '$', '%', '&', '=', '?', '@'])]
hotels.com: the `allowed` property for this rule can be shortened to: [Upper, Lower, Custom(['!', '#', '$', '%', '&', '(', ')', '*', '@', '^'])]
cvs.com: the `allowed` property for this rule can be shortened to: [Custom(['!', '#', '$', '%', '&', '(', ')', '*', '@', '^'])]
southwest.com: the `allowed` property for this rule can be shortened to: [Lower, Custom(['!', '#', '$', '%', '(', ')', '*', ',', '.', '/', ':', ';', '@', '\\', '^'])]
```

So for example, the rule for `marriot.com` is currently:

```
minlength: 8; maxlength: 20; required: lower; required: upper; required: digit; allowed: [abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789$!#&@?%=];
```

which the tool is saying can be shortened to:

```
minlength: 8; maxlength: 20; required: lower; required: upper; required: digit; allowed: [$!#&@?%=];
```

because all of the other characters in the custom character class are already part of the `required` classes in the form of `lower`, `upper`, or `digit`.

It would be great to shorten those `allowed` properties in this PR as well 😄 let me know if that makes sense!

rmondello:
@Cldfire I think you could directly edit the PR to make the additional changes you want, assuming @CasualDeveloper is okay with that. I’ve done this a few times to fix up tiny style things for people before landing their PR.

Also, I pushed another change to https://developer.apple.com/password-rules/. All I added was simple ordering of rules that are exactly `required: lower;`, `required: upper;`, `required: digit;`, and `required: special;`. All other rules, including custom character classes and `required: lower, upper;` (aka letter), are ordered in the order they appear. I’ll clean this up more later, but this should be enough to get most PRs that come in in a vaguely-matching order. :-)

Cldfire:
> @Cldfire I think you could directly edit the PR to make the additional changes you want, assuming @CasualDeveloper is okay with that. I’ve done this a few times to fix up tiny style things for people before landing their PR.

Good thought! Done in https://github.com/apple/password-manager-resources/pull/395/commits/3d0ae0746acd7df170ddc43b730c0e45c130ecb4.

> Also, I pushed another change to https://developer.apple.com/password-rules/. All I added was simple ordering of rules that are exactly `required: lower;`, `required: upper;`, `required: digit;`, and `required: special;`. All other rules, including custom character classes and `required: lower, upper;` (aka letter), are ordered in the order they appear. I’ll clean this up more later, but this should be enough to get most PRs that come in in a vaguely-matching order. :-)

Nice!

Cldfire (approved):
LGTM!

rmondello (approved):

CasualDeveloper:
Hi all, everything looks good here to me! Nice job, you two! Thanks for making one of my first PRs such a great experience! 💝

Issue #389, opened 2020-12-08
RFC: Add a new quirk representing "one-way" credential sharing


Cldfire:
This RFC proposes a new type of quirk that would make it possible to represent the concept of "one-way" credential sharing between sites. When compared to the existing shared credentials backends quirk, the new quirk would make it possible to express credential sharing with more precision in certain situations, such as domain redirects and SaaS login sites.

## Background

Our README has [some great background](https://github.com/apple/password-manager-resources#websites-with-shared-credential-backends) on why it's helpful for the repo to maintain a list of websites that share credential backends.

The shared credential backends quirk, as defined by the README, accepts entries that meet the following definition:

> The file quirks/websites-with-shared-credential-backends.json contains a list of groups of websites that share the same credential backend and serve pages where users can sign in, accepting accounts from the others. For example, adding first.website and second.website means that first.website and second.website each serve a page (e.g. first.website/login and second.website/login) where the same accounts are valid for signing in, despite the different domains. It wouldn’t be appropriate to associate google.com.il to google.com because google.com.il redirects to accounts.google.com for sign-in, and google.com.il never serves a login page.

Note that this definition excludes certain situations. I'd like to discuss these situations and the reasoning behind the proposed addition of a new type of quirk as opposed to expanding the scope of shared credential backends.

### Redirects

Redirects are domains that served login pages but were changed at some point in time to redirect to another domain where the same credentials are still valid.

A specific example of this is [discord.com](https://discord.com). [discordapp.com](https://discordapp.com) was [recently changed](https://support.discord.com/hc/en-us/articles/360042987951) to redirect to [discord.com](https://discord.com), a move that caused some confusion amongst users of password managers since logins for `discordapp.com` would not be suggested at `discord.com`.

Other past examples of such situations include [discussions.agilebits.com](https://discussions.agilebits.com) -> [1password.community](https://1password.community) and [me.com](https://me.com) -> [icloud.com](https://icloud.com).

The current definition of the shared credential backends quirk excludes these redirect scenarios, since the domain doing the redirecting doesn't itself serve a login page.

### SaaS Login Pages

Domains that initiate the login process but then redirect to a SaaS provider to complete authorization. Example: both [truecore.org](https://www.truecore.org/) and [sagaftrafcu.org](https://www.sagaftrafcu.org/) initiate authorization with a username and then redirect to [netteller.com](https://www.netteller.com) to complete the login process (and to provide the actual online service).

[...]

PR #398, opened 2020-12-16
Add Pinterest associated websites

MartinDevi:
### Overall Checklist

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -295,7 +295,6 @@
         "pinterest.fr",
         "pinterest.de",
         "pinterest.es",
-        "pin.it",
         "pinterest.com.au",
         "pinterest.se",
         "pinterest.ph",
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -289,6 +289,31 @@
         "optum.com"
     ],
     [
+        "pinterest.com",
+        "pinterest.ca",
+        "pinterest.co.uk",
+        "pinterest.fr",
+        "pinterest.de",
+        "pinterest.es",
+        "pin.it",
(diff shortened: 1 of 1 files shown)


rmondello:
These seem good, except pintrest.com, which is a redirect.

rmondello (changes requested):
Should we remove pintrest.com?

Cldfire:
The `pintrest.com` entry would fit very well into the format defined in https://github.com/apple/password-manager-resources/issues/389 😄

Cldfire:
Thinking about this some more, it might not be worth adding `pintrest.com` to the format proposed in #389.

We can assume that `pintrest.com` was registered by Pinterest to prevent someone else from buying up the (very similar) domain and using it for phishing. There's no reason why a user of a password manager should encounter this redirect during normal web browsing. In fact, should the user end up with credentials associated with the `pintrest.com` domain, I would argue that that's an issue that should be corrected by the user; a password manager shouldn't attempt to transparently adjust for that.

With that in mind, I would agree with @rmondello; `pintrest.com` should be removed here.

MartinDevi:
Removed `pintrest.com`

quirks/websites-with-shared-credential-backends.json
```diff
@@ -288,6 +288,30 @@
...
         "optum.com"
     ],
+    [
+        "pinterest.com",
+        "pinterest.ca",
+        "pinterest.co.uk",
+        "pinterest.fr",
+        "pinterest.de",
+        "pinterest.es",
+        "pin.it",
```
Cldfire, line 298: I'd say similar logic applies here; `pin.it` is a redirect and should be removed from this PR.

Cldfire:
Everything LGTM after removing `pin.it` :)

Cldfire (approved):

rmondello (approved):

PR #400, opened 2020-12-18
Add password rule for extraspace.com

eepskamp:
### Overall Checklist

#### for password-rules.json

extraspace.com doesn't accept a standard generated password from Safari.

The error message says a password should include an uppercase letter, a number, or a special character. The password in the picture has all three of those requirements, so that error message isn't telling the full story. 

I grabbed the validation regex for the password field from the HTML:
^(?=.*([A-Z]|\d|[$@$!%*#?&]))[A-Za-z\d$@$!%*#?&]{8,20}$

It looks like our generated password is being rejected because of the - character. The rule I added specifies allowed special characters for this site.

I tested this rule by downloading 10,000 rules from the Password Validation tool and applying the regex to each password in that file. All of the generated rules matched.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -215,6 +215,9 @@
     "expertflyer.com": {
         "password-rules": "minlength: 5; maxlength: 16; required: lower, upper; required: digit;"
     },
+    "extraspace.com": {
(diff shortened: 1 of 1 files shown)


rmondello (approved):

rmondello:
Thanks @eepskamp! Awesome, thorough contribution.

PR #401, opened 2020-12-22
Add Nord associated websites 

ChrisMarstaller:
### Overall Checklist

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -325,8 +325,10 @@
         "nsn.com"
     ],
     [
-        "nordvpn.com",
-        "nordaccount.com"
+        "nordaccount.com",
+        "nordlocker.com",        
+        "nordpass.com",
+        "nordvpn.com"       
     ]
     [
         "optumrx.com",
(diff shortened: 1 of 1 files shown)


rmondello:
All of the domains (sans nordaccount) redirect to nordaccount, so I don’t think it’s appropriate to incorporate this change as-is. @Cldfire

Cldfire:
> All of the domains (sans nordaccount) redirect to nordaccount, so I don’t think it’s appropriate to incorporate this change as-is. @Cldfire

Correct; this would be best suited as part of the format detailed in #389.

That being said, I'm unable to find a login form being served under a URL other than `https://my.nordaccount.com/login`, which means no entry should be required for these sites. @ChrisMarstaller, have you encountered a login for one of these accounts that is _not_ served under the `nordaccount.com` domain? If so, would you be able to share some steps to find the login form? 😄

ChrisMarstaller:
Apparently, my account was still on their old authentication method when I submitted the pull request, but I've since been upgraded to Nord Account and I'm seeing the same result as @rmondello. 

However, I noticed they are using both nordaccount.com and ndaccount.com (I did a password reset and it uses this domain) for authentication. The branch was updated with the domain changes.

ChrisMarstaller:
Since the domains are no longer being actively used, I'm closing this pull request and will revisit when a new format is available to handle historic data.

PR #402, opened 2020-12-26
Update websites-with-shared-credential-backends.json

AminIssa:
### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -6,6 +6,7 @@
     ],
     [
         "airfrance.com",
+        "klm.com",
         "flyingblue.com"
     ],
     [
@@ -293,10 +294,6 @@
         "kcls.org"
     ],
     [
-        "klm.com",
-        "flyingblue.com"
-    ],
-    [
         "lookmark.io",
         "lookmark.link"
     ],
@@ -367,7 +364,6 @@
      [
         "sfpl.bibliocommons.com",
         "sfpl.overdrive.com",
-        "sfpl.org"
     ],
     [
         "shopdisney.com",
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -252,6 +252,10 @@
         "messenger.com"
     ],
     [
+        "fandangonow.com",
+        "fandango.com"
+    ],
+    [
         "fnac.com",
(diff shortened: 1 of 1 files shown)


quirks/websites-with-shared-credential-backends.json
```diff
@@ -351,6 +363,11 @@
         "seattle.bibliocommons.com",
         "spl.overdrive.com",
         "spl.org"
+    ],
+     [
+        "sfpl.bibliocommons.com",
+        "sfpl.overdrive.com",
+        "sfpl.org"
```
Cldfire, line 370: Is there a login form hosted on `sfpl.org`? As far as I can tell it only redirects to `https://sfpl.bibliocommons.com/` for login, which wouldn't require an entry here.

quirks/websites-with-shared-credential-backends.json
```diff
@@ -288,6 +292,10 @@
         "kcls.overdrive.com",
         "kcls.org"
     ],
+    [
+        "klm.com",
+        "flyingblue.com"
```
Cldfire, line 297: Up at the top of this file we already have this entry:

```
    [
        "airfrance.com",
        "flyingblue.com"
    ],
```

We should add `klm.com` there rather than creating a new entry.

Cldfire:
There are some great additions here! Just a few thoughts which I've left above :)

quirks/websites-with-shared-credential-backends.json
AminIssa: My mistake, you are correct.

quirks/websites-with-shared-credential-backends.json
AminIssa: Fair point.

quirks/websites-with-shared-credential-backends.json
AminIssa: I'm new to GitHub, so forgive me for asking, but is there a way for me to fix this for your review?

Cldfire: No worries! You did the right thing :) Just needed a small tweak to the JSON syntax which I made in https://github.com/apple/password-manager-resources/pull/402/commits/3ca3323727cc998d1a60da9cf9946d5239f0feb8

Cldfire (approved):
Thank you!

PR #407, opened 2021-01-13
Add Liebherr

BastianZim:
### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -298,6 +298,10 @@
         "kcls.org"
     ],
     [
+        "liebherr.com",
+        "myliebherr.com"
+    ],
+    [
         "lookmark.io",
         "lookmark.link"
     ],


BastianZim:
Proof: 

- https://whois.domaintools.com/liebherr.com
- https://whois.domaintools.com/myliebherr.com

It's a bit of a weird quirk because registration is done on myliebherr.com and then login only on liebherr.com which then forwards to myliebherr.com.

Cldfire:
I clicked around a bit, but I wasn't able to find a login or registration page that wasn't under the domain `login.liebherr.com`, which makes it seem like this quirk shouldn't be necessary.

Do you happen to have a link to a login or registration page that's on the `myliebherr.com` domain?

BastianZim:
Hmm I noticed this quirk because both domains were showing up in the safari password manager so I think it's valid but let me sign up again and document everything so it's verifiable. Will do that just now.

BastianZim:
@Cldfire Found it again:

Registration: https://home.myliebherr.com/en/usa/my-profile/register
Login: https://home.liebherr.com/en/deu/home/homepage.html

At:

Screenshot 2021-01-17 at 13 05 46

Cldfire:
Ah, I see. Thanks for finding a link to that! 👍

Cldfire (approved):

PR #409, opened 2021-01-15
Add secure.orclinic.com

parkerbxyz:
This PR adds `secure.orclinic.com`.

Screen Shot 2021-01-14 at 8 42 27 PM

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -432,7 +432,7 @@
         "password-rules": "minlength: 8; maxlength: 12; required: lower, upper; required: digit; allowed: [-!#$%&'()*,.:;=?^{}];"
     },
     "secure.orclinic.com": {
-        "password-rules": "minlength: 6; maxlength: 15; required: lower, digit; allowed: ascii-printable;"
+        "password-rules": "minlength: 6; maxlength: 15; required: lower; required: digit; allowed: ascii-printable;"
     },
     "secure-arborfcu.org": {
         "password-rules": "minlength: 8; maxlength: 15; required: lower; required: upper; required: digit; required: [!#$%&'()+,.:?@[_`~]];"
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -431,6 +431,9 @@
     "santander.de": {
         "password-rules": "minlength: 8; maxlength: 12; required: lower, upper; required: digit; allowed: [-!#$%&'()*,.:;=?^{}];"
     },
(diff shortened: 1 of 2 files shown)


quirks/password-rules.json
```diff
@@ -431,6 +431,9 @@
     "santander.de": {
         "password-rules": "minlength: 8; maxlength: 12; required: lower, upper; required: digit; allowed: [-!#$%&'()*,.:;=?^{}];"
     },
+    "secure.orclinic.com": {
+        "password-rules": "minlength: 6; maxlength: 15; required: lower, digit; allowed: ascii-printable;"
```
parkerbxyz, line 435: Am I understanding the [documentation](https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules) correctly that if `allowed` is not specified, it defaults to `ascii-printable`? If so, would it be better to omit `allowed: ascii-printable` from this rule?

Cldfire: `allowed: ascii-printable` does indeed need to be included in this rule; otherwise, the only allowed character sets would be those that are `required` (`lower` + `digit`).

The `ascii-printable` default is only used if neither `required` nor `allowed` are specified.

Cldfire, line 435: Also, the `required` rule as it is currently written:

`required: lower, digit;`

isn't sufficient for the rule described in the screenshot from the site.

The site requires at least one lowercase letter _and_ at least one digit. The rule above, however, only requires one lowercase letter _or_ one digit.

This will represent the site's rules accurately:

`required: lower; required: digit;`

requiring both a lowercase letter and a digit to be present in the password.

parkerbxyz: Thanks, @Cldfire! I didn’t realize the distinction between writing the rule that way. I’ll get that fixed.

parkerbxyz:
```suggestion
        "password-rules": "minlength: 6; maxlength: 15; required: lower; required: digit; allowed: ascii-printable;"
```

parkerbxyz: Thanks for the clarification! 😊

rmondello (approved):

rmondello:
Looks good to me, but since @Cldfire was reviewing, I’ll let them merge.

Maitha990:
Better to marge them

ALMESMARI

Cldfire (approved):
Looks great! Thank you @parkerbxyz 😄

PR #413, opened 2021-01-22
Add password rules for hyresbostader.se

tomasf:
The [signup page](https://www.hyresbostader.se/User/register.aspx) says: "Mellan 6 och 20 tecken, bokstäver och siffror". Translation: "Between 6 and 20 characters, letters and digits". I have verified that the page does seem to require both letters and numbers, but doesn't accept things like dashes (which is used in Safari by default).

### Overall Checklist

#### for password-rules.json


Cldfire (approved):
Thank you! 🙂

harshmg121:
@Cldfire

PR #414, opened 2021-01-25
Add benefitslogin.discoverybenefits.com

parkerbxyz:
Screen Shot 2021-01-24 at 4 53 34 PM

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -65,6 +65,9 @@
     "battle.net": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower, upper; allowed: digit, special;"
     },
+    "benefitslogin.discoverybenefits.com": {
+        "password-rules": "minlength: 10; required: upper; required: digit; required: special; allowed: [!#$%&*?@];"
+    },
     "benjerry.com": {
         "password-rules": "required: upper; required: upper; required: digit; required: digit; required: special; required: special; allowed: lower;"
     },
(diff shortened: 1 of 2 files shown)


quirks/password-rules.json
```diff
@@ -65,6 +65,9 @@
     "battle.net": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower, upper; allowed: digit, special;"
     },
+    "benefitslogin.discoverybenefits.com": {
+        "password-rules": "minlength: 10; required: upper; required: digit; required: special; allowed: [!#$%&*?@];"
```
Cldfire, line 69: `required: special; allowed: [!#$%&*?@];` should be compacted down to `required: [!#$%&*?@];`. As it is currently written, this rule allows for any special character in the `special` character class (since the characters allowed by a password rule are the union of the `required` and `allowed` character classes), which isn't quite what we want here 🙂.

After that, we can add `allowed: lower;` to this rule, since the site allows for lower-case characters based on the screenshot.

Cldfire (changes requested):

quirks/password-rules.json
parkerbxyz: Thanks, @Cldfire! Does this look better?

```suggestion
        "password-rules": "minlength: 10; required: upper; required: digit; required: [!#$%&*?@]; allowed: lower;"
```

Cldfire: Yep, that looks great 😄

Cldfire (approved):

PR #419, opened 2021-01-30
Add password rule and change password URL for astonmartinf1.com

aaronraimist:
### Overall Checklist

#### for password-rules.json

screenshot


rmondello (changes requested):
Do they allow any special characters? I know the rule doesn’t mention them, but we could test manually.

aaronraimist:
Okay I tested some special characters. Everything I tried was allowed and I've added those to the rule.

rmondello:
Two characters I can spot missing are `-` and ` ` (space). At this point, might be better to put `allowed: special;`.

aaronraimist:
Yep those (and `{}`) worked as well so I switched over to `allowed: special;`

rmondello (approved):

PR #423, opened 2021-02-09
Add shared backend for Logitech's LOGI ID

ejdyksen:
Logitech's various properties share a single login system served from different domains, as documented [here](https://www.logitech.com/en-us/promo/logi-id.html).

Verified by signing up for a new account at id.logi.com (with a unique password), and was able to successfully login at all listed domains.

### Overall Checklist


Cldfire (approved):
Looks great, thank you!

PR #447, opened 2021-04-16
Fix failing linting for password rules sort order

bradcush:
The password rules for `guardiananytime`, `la-z-boy`, and `secure-arborfcu` we're all out of order causing the linting job to fail when branches were pushed and pull requests were opened.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -254,11 +254,11 @@
     "google.com": {
         "password-rules": "minlength: 8; allowed: lower, upper, digit, [-!\"#$%&'()*+,./:;<=>?@[^_{|}~]];"
     },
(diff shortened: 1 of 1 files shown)


rmondello (approved):

quirks/password-rules.json
```diff
@@ -254,12 +254,12 @@
     "google.com": {
         "password-rules": "minlength: 8; allowed: lower, upper, digit, [-!\"#$%&'()*+,./:;<=>?@[^_{|}~]];"
     },
-    "gwl.greatwestlife.com": {
-        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [-!#$%_=+<>];"
```
RmksRathnayaka, line 258: Change

PR #451, opened 2021-04-21
Add app.parkmobile.com

GeoJunkie:
### Overall Checklist

#### for password-rules.json

#### Site requirements

From their [sign-up page](https://app.parkmobile.io/register):

> Your password must contain 8-25 characters, 1 lowercase letter, 1 uppercase letter, 1 number, and 1 of the following special characters !@#$%^&

Screen Shot 2021-04-21 at 11 38 55

It also appears as a tooltip when changing your password.

Screen Shot 2021-04-21 at 11 39 47

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -86,6 +86,7 @@
     "nytimes.com": "https://myaccount.nytimes.com/seg/profile",
     "orcid.org": "https://orcid.org/account",
     "overleaf.com": "https://www.overleaf.com/user/settings",
+    "app.parkmobile.vo": "https://app.parkmobile.io/account/settings",
     "patreon.com": "https://www.patreon.com/settings/profile",
     "paypal.com": "https://www.paypal.com/myaccount/security/password/change",
     "pilotflyingj.com": "https://portal.pilotflyingj.com/myrewards/forgot-password",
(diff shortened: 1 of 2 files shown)


quirks/change-password-URLs.json
```diff
@@ -86,6 +86,7 @@
     "nytimes.com": "https://myaccount.nytimes.com/seg/profile",
     "orcid.org": "https://orcid.org/account",
     "overleaf.com": "https://www.overleaf.com/user/settings",
+    "app.parkmobile.vo": "https://app.parkmobile.io/account/settings",
```
rmondello, line 89: .vo?

rmondello (changes requested):

rmondello:
@GeoJunkie Thanks for the PR! Looks mostly good, but please check out the comment I left. Will be happy to approve this with a revision.

llevieux:
Was just about to submit this exact PR!  It is indeed app.parkmobile.io, contrary to the commit message and the json label.  

Until recently it used to be parkmobile.us (and dlweb.parkmobile.us), which no longer resolves.  Is it worth adding these three sites to websites-with-shared-credential-backends.json so that old passwords move over to the .io site? or is adding defunct/moved sites against policy?  

Thanks rmondello for this fantastic feature. Never contributed here so hope I'm doing this right.

quirks/change-password-URLs.json
GeoJunkie: Whoops! Fixed in 51d44b2c2db7a322e2a7638bac14509162edcf2f

GeoJunkie:
Sorry for the delay in revising this @rmondello . It should be ready now.

PR #397, opened 2020-12-16
Remove obsolete websites with shared credential backend

MartinDevi:
There are quite a few domains which don't match the criteria. As started in the project [README](https://github.com/apple/password-manager-resources#websites-with-shared-credential-backends):

> It wouldn’t be appropriate to associate google.com.il to google.com because google.com.il redirects to accounts.google.com for sign-in, and google.com.il never serves a login page.

I'm guessing that these domains matched the criteria when they were added but the website maintainers have changed the behavior since then. 

By that logic, the domains in this MR are removed for the following reasons.
- alltrails.io redirects to alltrails.com
- ameritrade.com redirects to tdameritrade.com
- anylistapp.com redirects to anylist.com
- capitalone360.com redirects to capitalone.com
- citibankonline.com and citibank.com redirect to citi.com
- comcast.net redirects to xfinity.com
- discordapp.com redirects to discord.com
- getdropbox.com redirects to dropbox.com
- microsoft.com and skype.com use live.com SSO
- office.com uses microsoftonline.com SSO
- mojang.com redirects to minecraft.net
- sonyentertainmentnetwork.com doesn't seem to exist anymore
- square.com redirects to squareup.com

- [z] I agree to the project's [Developer Certificate of Origin](https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md)
- [z] The top-level JSON objects are sorted alphabetically
- [z] There are no [open pull requests](https://github.com/apple/password-manager-resources/pulls) for the same update

PR #453, opened 2021-04-26
transferwise.com is now wise.com

irew:
More details: https://wise.com/help/articles/4LB2UIQjHBGiof7UBipZpU/how-will-the-change-to-wise-affect-my-login-account-details-and-card

### Overall Checklist

The service has changed their domain as explained here:
https://wise.com/gb/blog/world-meet-wise

### DIFF
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -539,6 +539,10 @@
         "luxilon.com"
     ],
     [
+        "wise.com",
+        "transferwise.com"
+    ],
+    [
         "worldlink.com.np",
         "nettv.com.np"
     ],


crleonard (approved):

Cldfire:
Hmm, technically speaking transferwise.com is simply a redirect to wise.com, so it's not a great fit for inclusion in `websites-with-shared-credential-backends.json`. See #397 and #389.

irew:
> Hmm, technically speaking transferwise.com is simply a redirect to wise.com, so it's not a great fit for inclusion in `websites-with-shared-credential-backends.json`. See #397 and #389.
> 
> That being said, no progress has happened on this front in a while, so I'm not necessarily opposed to merging this in the meantime.

You're right, however, this is a way that allows password managers to provide a better experience for those users who have saved a credential with the previous, and outdated URL. Though that is a mis-use of the original purpose. The risk is that the domain transferwise.com expires and changes ownership in future, however that also applies to any items in this list.

Curious to hear your thoughts on this one @rmondello

PR #470, opened 2021-06-07
Add password rule for lsacsso.b2clogin.com

mlinington:
A couple notes not included in the commit message:
- One option to satisfy the requirements is to encode the 3/4 requirement in six required blocks such that missing any two character classes will cause one of the requirements to fail. (e.g.: required: upper, lower; required: upper, digit; required: upper, special...). I opted against this since I'm pretty sure simplifying the requirements by always having an upper and a lower helps to preserve entropy. If you all have a different policy, let me know and I'll be happy to fix!
- Here's a screenshot of the requirement text:
image

Testing:
- Tested several sets of 10000 passwords from the password generator against the website's regex.
- Checked a couple passwords against the website. It's not a website I own so I'm not going to spam them with test accounts.

commit 40c68b3010a426e06ae4639cdc35190bab1188cf
Author: mlinington <mail@mlinington.com>
Date:   Sun Jun 6 20:47:23 2021 -0700

    Add password rule for lsacsso.b2clogin.com
    
    Law school admission council password requirements:
    > 8-16 characters, containing 3 out of 4 of the following: Lowercase
    > characters, uppercase characters, digits (0-9), and one or more of the
    > following symbols: @ # $ % ^ & * - _ ! . ?
    
    The js source looks to use the following regex to perform this
    validation:
    > ( (?=.*[a-z])(?=.*[A-Z])(?=.*\d)|
    >   (?=.*[a-z])(?=.*[A-Z])(?=.*[^A-Za-z0-9])|
    >   (?=.*[a-z])(?=.*\d)(?=.*[^A-Za-z0-9])|
    >   (?=.*[A-Z])(?=.*\d)(?=.*[^A-Za-z0-9])
    > ) ([A-Za-z\d@#$%^&*\-_+=[\]{}|\\:',?/`~"();!]|\.(?!@)){8,16}
    
    There are discrepancies between the regex and the text description of
    the requirements. The regex allows more special characters but disallows
    a "@" following a "." for some reason. The rule added by this commit
    uses only the characters in the textual description, minus the "." (to
    remove the possibility of a ".@" sequence).

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -363,7 +363,7 @@
         "password-rules": "minlength: 8; maxlength: 12; required: lower, upper; required: digit;"
     },
     "lsacsso.b2clogin.com": {
-        "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit, [!#$%&*-?@^_];"
+        "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit, [-!#$%&*?@^_];"
     },
     "lufthansa.com": {
         "password-rules": "minlength: 8; maxlength: 32; required: lower; required: upper; required: digit; required: [!#$%&()*+,./:;<>?@\"_];"
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -363,7 +363,7 @@
         "password-rules": "minlength: 8; maxlength: 12; required: lower, upper; required: digit;"
     },
     "lsacsso.b2clogin.com": {
-        "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit,[!#$%&*+?@^];"
+        "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit, [!#$%&*-?@^_];"
     },
     "lufthansa.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -362,6 +362,9 @@
     "lowes.com": {
         "password-rules": "minlength: 8; maxlength: 12; required: lower, upper; required: digit;"
     },
+    "lsacsso.b2clogin.com": {
+        "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit,[!#$%&*+?@^];"
```
Cldfire, line 366: The commit description states:

> The rule added by this commit uses only the characters in the textual description, minus the "." (to remove the possibility of a ".@" sequence).

However, the password rule as written here allows for the `+` character, which is not in the textual description in the screenshot. Should we remove `+` from the rule?

mlinington: You're right! I think this must've been a mis-copy. Thanks for catching!

I've updated the character set to remove '+' and add '-' and '_' as special characters.

Cldfire:
Thank you for the PR! I made a quick commit to move the `-` to the front of the special character class as it has to be located there in order for the rule to parse correctly.

Also sorry for the delay on replying here 🕐

Cldfire (approved):

PR #473, opened 2021-06-09
Add password rules for flysas.com 

vdbelt:
### Overall Checklist

- [x] There's [evidence](https://en.wikipedia.org/wiki/Scandinavian_Airlines) the domains are related (SSL certificates, DNS entries, valid links between sites, legal documents etc.)
- [x] The new group serves login pages on each of the included domains, and those login page accept accounts from the others.

#### for password-rules.json

Screenshot 2021-06-09 at 02 17 51

Note that when filling a password, the green check mark for special characters will show up even if using ;*/. It however doesn't actually allow you to save the password.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -252,7 +252,7 @@
         "password-rules": "minlength: 8; max-consecutive: 3; required: lower; required: upper; required: digit; allowed: [-!@#$%^&*_+=`|(){}[:;,.?]];"
     },
     "flysas.com": {
-        "password-rules": "minlength: 8; maxlength: 14; required: lower; required: upper; required: digit; required: [-(~!@#$%^&_+=`|(){}[:\"'<>,.?]];"
+        "password-rules": "minlength: 8; maxlength: 14; required: lower; required: upper; required: digit; required: [-~!@#$%^&_+=`|(){}[:\"'<>,.?]];"
     },
     "fnac.com": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit;"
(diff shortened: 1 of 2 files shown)


quirks/password-rules.json
```diff
@@ -251,6 +251,9 @@
     "fedex.com": {
         "password-rules": "minlength: 8; max-consecutive: 3; required: lower; required: upper; required: digit; allowed: [-!@#$%^&*_+=`|(){}[:;,.?]];"
     },
+    "flysas.com": {
+        "password-rules": "minlength: 8; maxlength: 14; required: lower; required: upper; required: digit; required: [(-~!@#$%^&_+=`|(){}[:\"'<>,.?]];"
```
rmondello, line 255: In a character class, `-` needs to be the first character, if present.

rmondello (changes requested):

quirks/password-rules.json
vdbelt: Thank you, fixed!

Cldfire (approved):

quirks/password-rules.json
```diff
@@ -251,6 +251,9 @@
     "fedex.com": {
         "password-rules": "minlength: 8; max-consecutive: 3; required: lower; required: upper; required: digit; allowed: [-!@#$%^&*_+=`|(){}[:;,.?]];"
     },
+    "flysas.com": {
+        "password-rules": "minlength: 8; maxlength: 14; required: lower; required: upper; required: digit; required: [-(~!@#$%^&_+=`|(){}[:\"'<>,.?]];"
```
rmondello, line 255: Looks like `(` is in the character class more than once.

rmondello (changes requested):

rmondello (approved):

PR #481, opened 2021-06-24
Add ae.com - American Eagle Outfitters

d19dotca:
Rules from American Eagle when changing a user password: "Passwords must be between 8-25 characters including one letter and one number and cannot match your previous three passwords."

### Overall Checklist

#### for password-rules.json


rmondello:
Do they accept any "special" characters?

d19dotca:
> Do they accept any "special" characters?

Yes, but it's not clear which ones specifically they accept. This is my first commit, is there a property I should have added that indicates it accepts special characters but not which ones specifically?

rmondello (approved):

Issue #486, opened 2021-07-02
americanexpress.com rejects passwords with the '-' character


geoffreygaren:
I see that we have a rule for americanexpress.com already:

"americanexpress.com": {

        "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 4; required: lower, upper; required: digit; allowed: [-%&_?#=];"

    },

But americanexpress.com rejects the password that Safari generates on Big Sur and latest beta.

I think the incompatibility is the '-' character. See attached error message image.
![Screen Shot 2021-07-02 at 11 15 45 AM](https://user-images.githubusercontent.com/6519785/124319203-be24e680-db2e-11eb-894b-4bcc81948c76.png)

It seems like the minimal fix would be to remove '-' from the allowed list. But maybe we should go further and remove all special characters, in case americanexpress.com changes again?

rmondello:
I think we can just remove `-` for now. @geoffreygaren, you up for making that change? :)

PR #489, opened 2021-07-08
Remove - special character americanexpress.com

billycastelli:
Fix https://github.com/apple/password-manager-resources/issues/486

#### for password-rules.json


rmondello:
Thanks @billycastelli!

rmondello (approved):

PR #498, opened 2021-07-26
Add password rules for Renaud-Bray

Xavientois:
When testing password generation with 1Password, there are no indications of the allowed characters until after the password is rejected by the site. Only after trying a generated password does an error message appear saying that special characters and accented characters are not allowed:

Screen Shot 2021-07-26 at 12 19 21 PM
Screen Shot 2021-07-26 at 12 19 27 PM

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -500,6 +500,9 @@
     "rejsekort.dk": {
         "password-rules": "minlength: 7; maxlength: 15; required: lower; required: upper; required: digit;"
     },
+    "renaud-bray.com": {
+        "password-rules": "minlength: 8; maxlength: 38; allowed: upper, lower, digits;"
+    },
     "ring.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -500,6 +500,9 @@
     "rejsekort.dk": {
         "password-rules": "minlength: 7; maxlength: 15; required: lower; required: upper; required: digit;"
     },
+    "renaud-bray.com": {
+        "password-rules": "minlength: 8; maxlength: 38; allowed: upper, lower, digits;"
```
rmondello, line 504: This isn’t a valid rule. Please use https://developer.apple.com/password-rules/ to craft a rule.

rmondello (changes requested):

rmondello (approved):

PR #501, opened 2021-08-04
Password rules and change password url for carsandbids.com

robotmlg:
### Overall Checklist

#### for password-rules.json

Screenshot:
Screen Shot 2021-08-04 at 13 17 24

Also, I am a developer at CarsAndBids.com, and I can confirm these requirements are correct.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -119,6 +119,9 @@
     "cardbenefitservices.com": {
         "password-rules": "minlength: 7; maxlength: 100; required: lower, upper; required: digit;"
     },
+    "carsandbids.com": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%&'()*+./:=?@[\^_`{}]];"
+    }
     "cb2.com": {
         "password-rules": "minlength: 7; maxlength: 18; required: lower, upper; required: digit;"
     },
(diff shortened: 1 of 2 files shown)


quirks/password-rules.json
```diff
@@ -119,6 +119,9 @@
     "cardbenefitservices.com": {
         "password-rules": "minlength: 7; maxlength: 100; required: lower, upper; required: digit;"
     },
+    "carsandbids.com": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%&'()*+./:=?@[\^_`{}]];"
+    }
```
rmondello, line 124: Missing the trailing comma here, making this JSON invalid.

rmondello (changes requested):
If you’re a developer at CarsAndBids.com, you can enhance your website without contributing to this repository!

For password rules, you can use the passwordRules attribute on your text fields. https://developer.apple.com/documentation/security/password_autofill/customizing_password_autofill_rules?language=objc

For a discoverable change password URL, you can use the A Well-Known URL for Changing Passwords. https://github.com/w3c/webappsec-change-password-url

That way, all browsers and password managers that adopt these standards will benefit. What do you think? Happy to discuss?

robotmlg:
Thanks Ricky, we've actually updated our password rules to be more flexible since I opened this PR, so they no longer require specification, We'll work on getting the well-known path implemented.  Thanks for the pointers.

PR #508, opened 2021-08-17
Add rules for major HK banks

tsekityam:
### Overall Checklist

#### for password-rules.json

#### [hsbc.com.hk](https://www.ebanking.hsbc.com.hk/1/2/faq/popsecadpwd)

> The secret word or code you will use to access your Internet banking service. You will specify you own password. Your password should be created in 6-30 characters. You may use a combination of numbers (0-9), and/or upper and lower case letters (A-Z, a-z), and/or special characters such as (@), underscore (_), hyphen (-), apostrophe (') and period (.). Other special characters will not be accepted.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -105,7 +105,7 @@
         "password-rules": "allowed: lower, upper, digit, [!$#@];"
     },
     "bochk.com": {
-        "password-rules": "minlength: 8; maxlength: 12; required: lower; required: upper; required: digit; allowed: [#$%&()*+,.:;<=>?@_];"
+        "password-rules": "minlength: 8; maxlength: 12; max-consecutive: 3; required: lower; required: upper; required: digit; allowed: [#$%&()*+,.:;<=>?@_];"
     },
     "box.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -104,6 +104,9 @@
     "bluesguitarunleashed.com": {
         "password-rules": "allowed: lower, upper, digit, [!$#@];"
     },
+    "bochk.com": {
+        "password-rules": "minlength: 8; maxlength: 12; required: lower; required: upper; required: digit; allowed: [#$%&()*+,.:;<=>?@_];"
```
rmondello, line 108: You can tack a `max-consecutive: 3;` onto this.

rmondello (approved):

rmondello:
Thanks for the contribution, @tsekityam! I can merge this, but I left one comment of an optional improvement you can make. Let me know whether you want to make it or not!

quirks/password-rules.json
tsekityam: Added `max-consecutive: 3;` to BOCHK rule

rmondello (approved):

rmondello:
Thanks @tsekityam!

PR #511, opened 2021-09-05
Add shared credentials for US Sailing and US Powerboating

wstyres:
Added shared credentials between ussailing.org and uspowerboating.com. Each site has its own login page but credentials are shared between the two sites. This can be seen best on [uspowerboating.com's login page](https://www1.uspowerboating.com/user/Login.aspx?cmsReturn=autoclose.aspx) where clicking the "Forgot member ID" or "Forgot password" links will take you back to ussailing.org.

### Overall Checklist

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -160,8 +160,8 @@
     },
     {
         "shared": [
-            "ussailing.org",
-            "uspowerboating.com"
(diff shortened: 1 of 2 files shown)


rmondello (approved):

rmondello:
Thanks @wstyres!

PR #520, opened 2021-09-25
IQVIA shared login on domains

karisair:
Both iqvia.com and iqvia-study-hub.herokuapp.com are using the same login details. Both logins are controlled by IQVIA.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -139,6 +139,12 @@
         ]
     },
     {
+        "shared": [
+            "iqvia.com",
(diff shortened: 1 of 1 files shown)


rmondello:
iqvia-study-hub.herokuapp.com doesn’t seem to be up right now. @karisair, what are you hoping to achieve with this change?

karisair:
Sorry, I'll close this one off as it's now changed to another subdomain on the Iqvia.com domain!

rmondello:
@karisair No need to be sorry! Things change! :)

PR #526, opened 2021-10-04
Add blade

BastianZim:
### Overall Checklist
- [ ] There are no [open pull requests](https://github.com/apple/password-manager-resources/pulls) for the same update

#### for password-rules.json

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -79,6 +79,15 @@
     },
     {
         "from": [
+            "flyblade.com"
+            ],
+        "to": [
+            "blade.com"
+            ],
+        "fromDomainsAreObsoleted": true
+    }
+    {
+        "from": [
             "discordapp.com"
         ],
         "to": [


quirks/shared-credentials.json
```diff
@@ -77,6 +77,15 @@
             "airnewzealand.com.au"
         ]
     },
+    {
+        "from": [
+            "flyblade.com"
+            ],
```
bradcush, line 83:
```suggestion
        ],
```

quirks/shared-credentials.json
```diff
@@ -77,6 +77,15 @@
...
+        "to": [
+            "blade.com"
+            ],
```
bradcush, line 86:
```suggestion
        ],
```

bradcush (changes requested):

bradcush:
Thanks for the submission @BastianZim! Small nit around the formatting and I can get this merged. Also if you wouldn't mind editing the description to remove any parts of the checklist that aren't relevant to your PR that would be great!

BastianZim:
Thanks for the review @bradcush Should all be fixed now.

PR #529, opened 2021-10-11
Moved centralfcu from historical to current shared credentials

ezfe:
### Overall Checklist

Proof of shared credentials:

Entering a correct password on https://www.centralfcu.com goes directly to the account page. However, entering *incorrect* credentials on https://www.centralfcu.com loads a login page on https://www.centralfcu.org/tob/live/usp-core/app/initialLogin. This establishes a relationship from the `.com` to the `.org`.

Navigating to https://www.centralfcu.org and clicking the logo will redirect to the `.com` version. This establishes the reverse relationship.


bradcush:
@ezfe Thanks for the submission! Just to clarify, is there a difference between logging in to `centralfcu.com` versus `centralfcu.org`, meaning the site that you have access to is actually different once you've gained access?

quirks/shared-credentials-historical.json
```diff
@@ -153,12 +153,6 @@
             "asiamiles.com"
         ]
     },
-    {
-        "shared": [
```
bradcush, line 157: Specify using `from`/`to` keys, the `to` being `centralfcu.org`

ezfe:
@rmondello Can you confirm whether `from`/`to` should be used when the login page appears on both domains, but always redirects *after* login to the second domain?

rmondello:
> @rmondello Can you confirm whether `from`/`to` should be used when the login page appears on both domains, but always redirects _after_ login to the second domain?

That's not the meaning. It's for when a domain redirects to another domain to do its login.

ezfe:
Thanks - I will resolve the conflict but then this is ready to merge

ezfe:
@bradcush Please re-review

bradcush (approved):

bradcush:
> Thanks - I will resolve the conflict but then this is ready to merge

Sorry for the confusion here, thanks for clarifying @rmondello!

PR #540, opened 2021-11-09
Added Password Rule for belkin.com

MatthewFolbigg:
Belkin have the following rules when creating a password for their retail/support website causing iCloud strong passwords to be rejected:

- Contain at least 1 number.
- Contain at least 1 letter.
- Contain at least 1 special character from the following $ ! @ ~ _ , % &
- Be at least 8 characters in length
- Not contain first name, last name, email address or spaces

belkin com Screenshot 2021-11-09

### Inconsistency in requirements help pop up
Each requirement will turn green once it is met in the password field. However it is worth noting the special characters requirement will display green when including some other special characters such as ^ or #. Although the requirement is displayed as met, it will still be rejected when submitting the form. 

Some generated passwords would not trigger the 'Not contain first/last name' requirement to show as met. However these passwords are still accepted when submitting the form.

### Overall Checklist

#### for password-rules.json


quirks/password-rules.json
```diff
@@ -80,6 +80,9 @@
     "bcassessment.ca": {
         "password-rules": "minlength: 8; maxlength: 14;"
     },
+    "belkin.com": {
+        "password-rules": "minlength: 8; required: lower, upper; required: digit; required: [$!@~_,%&];"
```
rmondello, line 84: You’ve left out `-`, which needs to be the first character in a class that it’s in. So: `required: [-$!@~_,%&];`.

You cited iCloud Keychain as not being compatible. These rules seem to be compatible to me.

rmondello (changes requested):

quirks/password-rules.json
MatthewFolbigg: Hey Ricky, 

"-" doesn't actually count towards the "Password must contain at least one of the following special characters:". This is why my iCloud generated passwords are rejected as the only special character used for mine is a "-".

I had originally assumed what looked like a "-" in the requirements pop up was actually a "~" so I replaced it in the rule. However looking back since there are other accepted characters omitted from the pop up, the inclusion of "~" is probably a mistake.

Including the "-" would break this rule in the event it is the only special character used. 

Below is a screenshot of an iCloud generated password failing. Even ignoring the requirements pop up, submitting it will reject the password.
Screenshot 2021-11-09 at 21 06 29

In this instance would it be preferable to include any characters tested and working in the rule, or only those suggested in the pop up that are also tested and working?

Thanks!

rmondello: Wow, tricky!

rmondello (approved):

PR #541, opened 2021-11-10
Add Amazon shared credential

N0chteil:
### Overall Checklist

_Domains according to [Wikipedia](https://en.wikipedia.org/wiki/Amazon_(company)#Website)_
I've checked all domains and all are working.


N0chteil:
(@rmondello)

PR #555, opened 2021-12-09
Update shared-credentials-historical.json

edouardroger:


rmondello:
Hi @edouardroger! Can you add the standard template information to your description?

edouardroger:
> Hi @edouardroger! Can you add the standard template information to your description?

Thank you for your answer. However, I guess I was wrong to submit this PR regarding the following provision 🤦‍♂️: 

> If using `shared`, the new group serves login pages on each of the included domains, and those login pages accept accounts from the others. (For example, we wouldn't use a `shared` association from `google.co.il` to `google.com`, because `google.co.il` redirects to `accounts.google.com` for sign in.)

- cfspart.impots.gouv.fr is the French tax office login page.
- FranceConnect, based on OpenID Connect, allows French users to sign in to other websites using their tax credentials through idp.impots.gouv.fr.

Apologies !

PR #567, opened 2022-01-17
Add change password URL for shutterfly

TylerCarberry:
### Overall Checklist


bradcush:
Confirming that this change URL redirects logged out users to a login form which subsequently redirects to the `account-settings` after a successful login. This page contains a link to the change password form, which can't be linked to more directly. Thanks for the submission @TylerCarberry!

bradcush (approved):

PR #576, opened 2022-02-02
Add seek domains to shared credentials

ashton-seek:
### Overall Checklist

Evidence:
If you go to https://www.seek.com.au and click 'Sign In' or 'Register' you will be taken to https://login.seek.com
If you go to https://www.seek.co.nz and click 'Sign In' or 'Register' you will be taken to https://login.seek.com
You can also register on AU and then sign in on NZ, they share the same credential backend.

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -285,6 +285,16 @@
     },
     {
         "from": [
+            "www.seek.com.au",
+            "www.seek.co.nz"
+        ],
+        "to": [
+            "login.seek.com"
+        ],
+        "fromDomainsAreObsoleted": true
+    },
+    {
+        "from": [
             "www.vistaprint.ca"
         ],
         "to": [


merijndejonge (approved):

merijndejonge (approved):

ashton-seek:
I've rebased this and sorted it based on the first domain in the object. Let me know if this is OK @rmondello (sorry for the nudge)

quirks/shared-credentials.json
```diff
@@ -255,6 +255,16 @@
             "wikimania.wikimedia.org"
         ]
     },
+    {
+        "from": [
+            "www.seek.com.au",
```
bradcush, line 260: Just being a little nit picky here but would you mind sorting it based on the actual website domain in this case, according to `seek`, as the `www` can be typical for many sites. Happy to merge after that minor change, thanks!

ashton-seek: That's fair, thanks. I've rebased and addressed this.

bradcush (approved):

quirks/shared-credentials.json
bradcush: I apologize 😬, I gave you the wrong advice, the order I proposed is preventing it from passing the shared credentials sort order linting which I believe you had right the first time.

ashton-seek: Oh 😄 I liked it better too. Ran the lint script itself and it sorted it for me, should be good now

PR #602, opened 2022-04-18
Add password rule for museumofflight.org

### LINKED ISSUE #586: Generated password does not fit the requirements of MuseumOfFlight .org
Hi,

When I use Apple's password manager to create a new password for this website, it fails. The generated password is 20 chars long. However, the required password needs to be between 8 to 15 chars long. 

This creates a failed attempt to complete the signup and requires use to manually create a password (which is probably awful a lot less safer).

Website link: MuseumOfFlight.org 
Attached a screenshot below.

Cheers, team! You rock!

Goktug

Screen Shot 2022-02-13 at 10 14 44 AM

tsekityam:
Fix #586

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -476,6 +476,9 @@
     "mpv.tickets.com": {
         "password-rules": "minlength: 8; maxlength: 15; required: lower; required: upper; required: digit;"
     },
+    "museumofflight.org": {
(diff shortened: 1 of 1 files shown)


aabosh (approved):

aabosh:
Looks good! Thanks!

PR #605, opened 2022-04-30
Update Citi password rules

xiaoxu193:
### Overall Checklist

#### for password-rules.json

From Citi bank's password change page:
```
Password Guidelines

    Different from your User ID
    No spaces
    No more than 2 consecutive identical characters
    Include between 8-64 characters
    No digital images or icons

Must Include:

    Include at least 1 number
    Include at least 1 uppercase letter
    Include at least 1 lowercase letter
    Include at least 1 special characters: ~ ` ! @ # $ % ^ & * ( ) _ - \ / |
```
image

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -150,7 +150,7 @@
         "password-rules": "minlength: 8; maxlength: 12; required: lower; required: upper; required: digit;"
     },
     "citi.com": {
-        "password-rules": "minlength: 6; maxlength: 50; max-consecutive: 2; required: lower, upper; required: digit; allowed: [_!@$]"
+        "password-rules": "minlength: 8; maxlength: 64; max-consecutive: 2; required: lower, upper; required: digit; required: [~`!@#$%^&*()_-\/|];"
     },
     "claimlookup.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [@#$%^&+=!];"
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -150,7 +150,7 @@
         "password-rules": "minlength: 8; maxlength: 12; required: lower; required: upper; required: digit;"
     },
     "citi.com": {
-        "password-rules": "minlength: 6; maxlength: 50; max-consecutive: 2; required: lower, upper; required: digit; allowed: [_!@$]"
+        "password-rules": "minlength: 8; maxlength: 64; max-consecutive: 2; required: lower, upper; required: digit; required: [~`!@#$%^&*()_-\/|];"
```
rmondello, line 153: This line should be:

```
"password-rules": "minlength: 8; maxlength: 64; max-consecutive: 2; required: lower, upper; required: digit; required: [-~`!@#$%^&*()_\/|];"
```

There's a subtle language quirk with character classes where, if `-` is present, it needs to be the first item in the class to take effect.

rmondello (changes requested):

quirks/password-rules.json
xiaoxu193: Ok got it. I have made the updates

bradcush (approved):

Deloww (commented):
🙏 thanks

rmondello (approved):

rmondello:
It looks like someone already landed an updated rule here. I don't think this PR is needed. If I'm wrong, please re-open!

PR #619, opened 2022-06-27
Add plazapremiumlounge.com password-rules.json

michaelabon:
Creating an account at plazapremiumlounge.com has very specific rules about which special characters are allowed. None of them are required.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -536,6 +536,9 @@
     "planetary.org": {
         "password-rules": "minlength: 5; maxlength: 20; required: lower; required: upper; required: digit; allowed: ascii-printable;"
     },
+    "plazapremiumlounge.com": {
+        "password-rules": "minlength: 8; maxlength: 15; required: lower; required: upper; required: digit; allowed: [!#$%&*,@^];"
+    },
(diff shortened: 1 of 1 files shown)


michaelabon:
I tested these password rules by hand, manually editing the password on the account creation form to see what would trigger validation errors. The only actual error message that will appear is:

Screenshot of the registration page for PlazaPremiumLounge.com, with the "Create Password" field in an error state with the error message of "Password must contain at least 1 lower case,1 upper case,1 numeric and 8 to 15 length character."

I used the Password Rules Validation tool to both create the string included in the PR and then to test several generated passwords.

Sameesunkaria:
Hi @michaelabon 👋 

Thank you for your contribution. Just wanted to additional context here; I had a chance to take a look at the password validation for the website. They seem to use the following regex:

```regex
/^(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z])[0-9a-zA-Z!@,#$%^&*]{8,15}$/
```

Your changes look good to me! :)

aabosh (approved):

aabosh:
Thank you @michaelabon!

PR #632, opened 2022-09-20
Add pretendo.network Shared Password Backend

InternalLoss:
### Overall Checklist

#### for password-rules.json

- [X] There's evidence the domains are currently related (SSL certificates, DNS entries, valid links between sites, legal documents etc.)
- [X] If using `shared`, the new group serves login pages on each of the included domains, and those login pages accept accounts from the others. (For example, we wouldn't use a `shared` association from `google.co.il` to `google.com`, because `google.co.il` redirects to `accounts.google.com` for sign in.)

As a developer of @PretendoNetwork, I can confirm that https://juxt.pretendo.cc/login uses the same credential backend as https://pretendo.network/account/login

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -269,6 +269,12 @@
         ]
     },
     {
+        "shared": [
+            "pretendo.network",
+            "pretendo.cc"
+            ]
+    },
+    {
         "from": [
             "tvnow.de",
             "tvnow.at",
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -272,7 +272,7 @@
         "shared": [
             "pretendo.network",
             "pretendo.cc"
-            ]
+        ]
     },
     {
         "from": [


quirks/shared-credentials.json
```diff
@@ -268,6 +268,12 @@
             "postnl.be"
         ]
     },
+    {
+        "shared": [
+            "pretendo.network",
+            "pretendo.cc"
+            ]
```
aabosh, line 275: Can you shift the closing bracket back a few spaces/tab so it's aligned with line 272? You can look at the shared group above yours as an example.

InternalLoss: Resolved in commit https://github.com/apple/password-manager-resources/pull/632/commits/9b4b03c3f97ab70b98060c45e4a254a7ce9e4945

aabosh (approved):

aabosh:
Thanks for your contribution! :)

PR #645, opened 2022-11-13
Remove minecraft.com password rules

ezfe:
### Overall Checklist

There are two reasons for removing this:

- `minecraft.com` redirects to `minecraft.net`, so this is the wrong domain anyways
- `minecraft.net` profiles must be migrated to `microsoft.com` profiles, so it is no longer possible to set a password on this domain

I have also removed the password changing URLs for `mojang.com` and `minecraft.net` because it is no longer possible to change those account passwords. They must be migrated to `microsoft.net` accounts.

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -80,8 +80,6 @@
     "meliuz.com.br": "https://www.meliuz.com.br/minha-conta/meus-dados/senha",
     "messagebird.com": "https://dashboard.messagebird.com/account/security",
     "microsoft.com": "https://account.live.com/password/Change",
-    "minecraft.net": "https://www.minecraft.net/profile",
-    "mojang.com": "https://www.minecraft.net/profile",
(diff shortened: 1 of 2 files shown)


Sameesunkaria:
Hi @ezfe!

The changes to the password rules look good! 
Even though you are no longer able to change passwords on `mojang.com` and `minecraft.net`, do you think directing the **existing** users to the homepage of those websites would provide a better user experience that directing them to a dedicated page for managing their account?

ezfe:
If users still have Mojang or Minecraft credentials then the best course of action is probably to send them to Minecraft.net where they can try to go through the login flow and it will send them to the right place from there.

there’s no account specific pages they’d be able to access since the credentials aren’t valid anymore.

Sameesunkaria (approved):
Sorry that it took a while to get to it but thank you for the contribution @ezfe! 🙂

PR #655, opened 2022-12-22
Add authenticate.seek domains to shared credentials

ashton-seek:
### Overall Checklist

`talent.seek.com.au` is now using `authenticate.seek.com` for authentication, it previously used `authenticate.seek.com.au`.
The old domain does not redirect to the new domain, as it is now obsolete and no longer linked to. But you can see that this assetlinks file has a similar association to the one suggested in this PR https://authenticate.seek.com/.well-known/assetlinks.json

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -108,6 +108,15 @@
         "fromDomainsAreObsoleted": true
     },
     {
+        "from": [
+            "authenticate.seek.com.au"
+        ],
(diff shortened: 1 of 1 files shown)


dmmaslenn:
Sorry that it took us some time to review this PR!
Looks like proposed change is not valid since authenticate.seek.com.au is obsolete. Please feel free to open new PR for shared-credentials-historical.json.

Issue #658, opened 2023-01-08
Add password rules for member.everbridge.net


ddkilzer:
Add password rules for member.everbridge.net.

This website hosts accounts for multiple organizations, many of which are used for emergency alert systems.

Alert SCC: https://member.everbridge.net/453003085615038/new
Sacramento Region Emergency Notification System: https://member.everbridge.net/453003085611271/new

However, a pseudo-random sampling from top Duck-Duck-Go search results for "[member.everbridge.net login](https://duckduckgo.com/?q=member.everbridge.net+login&ia=web)" shows that they all use the same password rules:

https://member.everbridge.net/892807736723690/new
https://member.everbridge.net/453003085617772/new
https://member.everbridge.net/1772417038942730/new
https://member.everbridge.net/453003085611768/new
https://member.everbridge.net/565703027458110/new
https://member.everbridge.net/892807736721501/new
https://member.everbridge.net/index/892807736729003#/signup
https://member.everbridge.net/892807736722469/new

![AlertSCC-1](https://user-images.githubusercontent.com/263571/211178768-41d684d1-dee5-4803-b98f-2e3756a22750.png)
![AlertSCC-2](https://user-images.githubusercontent.com/263571/211178771-31748536-c1f6-4e89-b642-5042e7e42780.png)

PR #659, opened 2023-01-08
Add password-rules for member.everbridge.net (#658)

ddkilzer:
This one website hosts accounts for many different organizations, but they all have the same password rules based on a pseudo-random sampling of "Sign Up" pages.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -479,6 +479,9 @@
     "medicare.gov": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [@!$%^*()];"
     },
+    "member.everbridge.net": {
+        "password-rules": "minlength: 8; required: lower, upper; required: digit; allowed: [!@#$%^&*()];"
+    },
     "metlife.com": {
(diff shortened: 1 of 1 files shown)


ddkilzer:
BTW, looking at <https://developer.apple.com/password-rules/>, I find it strange that none of the "`allowed:`" characters are ever used if they're not required for this rule:

    "minlength: 8; required: lower, upper; required: digit; allowed: [!@#$%^&*()];"

rmondello:
> BTW, looking at https://developer.apple.com/password-rules/, I find it strange that none of the "`allowed:`" characters are ever used if they're not required for this rule:
> 
> ```
> "minlength: 8; required: lower, upper; required: digit; allowed: [!@#$%^&*()];"
> ```

The generator on https://developer.apple.com/password-rules/ is opinionated, just like iCloud Keychain is. It doesn’t think that any of those characters make for a good password, and avoids using them where it can.

We should probably add a toggle between a more “pure” generation and an iCloud Keychain-style generation.

rmondello (approved):

PR #660, opened 2023-01-09
Add password rule for Boston Public Library Bibliocommons PIN

adinklotz:
### Overall Checklist

#### for password-rules.json

The Boston Public Library's online portal requires a 4-digit PIN. Annoyingly, although there are a [ton](https://www.bibliocommons.com/about/libraries/united-states) of libraries at {libraryname}.bibliocommons.com, from clicking through to a couple of different login pages it appears that some require PINs and some allow real passwords, meaning we can't just add a blanket rule for bibliocommons.com. I added BPL here since that's the one I encountered, but for completeness we'd have to go through that whole list and check each library's rules individually. Ugh.

For reference, here's the message on the change password page:
Screenshot 2023-01-09 at 10 37 25 AM

It's not clear from that what 'characters' entails, but it complains if you include non-digits:
Screenshot 2023-01-09 at 10 44 08 AM

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -134,6 +134,9 @@
     "brighthorizons.com": {
         "password-rules": "minlength: 8; maxlength: 16;"
     },
+    "bpl.bibliocommons.com": {
+        "password-rules": "minlength: 4; maxlength: 4; required: digit;"
+    }
     "callofduty.com": {
         "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 2; required: lower, upper; required: digit;"
     },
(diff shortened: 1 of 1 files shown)


rmondello:
@adinklotz I think we need to deploy `exact-domain-match-only` in this PR. (See the project README for more information!)

adinklotz:
@rmondello Do we? The default behavior seems like what we want:
> A rule for a.example.com will match URLs on a.example.com as well as *.a.example.com, but will not match other subdomains of example.com such as b.example.com

so we get (\*.)bpl.bibliocommons, but other .bibliocommons will be separate. (happy to add it, just not clear to me that it's warranted)

rmondello (approved):

PR #661, opened 2023-01-09
Add readymag.com to password-rules and change-password-URLs

esafev:
### Overall Checklist

#### for password-rules.json

#### Screenshot with password rules:
image


rmondello:
Hi @esafev! Thanks for the contribution! Were you able to test that the website accepts all `special` characters?

esafev:
Yes, any character that is not a unicode alphabetic and is not a digit can be used on this site.🖖🏻

rmondello (approved):

PR #679, opened 2023-04-19
Added shared credential for nelnet

Tschrock:
`nelnet.com` and `secure.nelnet.net` serve the same site, including login and account pages, and you can log into your account on either domain. Note, the bare domain `nelnet.net` redirects to `nelnet.com`. 

https://nelnet.com/welcome
https://secure.nelnet.net/welcome

### Overall Checklist

- [X] There's evidence the domains are currently related (SSL certificates, DNS entries, valid links between sites, legal documents etc.)
- [X] If using `shared`, the new group serves login pages on each of the included domains, and those login pages accept accounts from the others. (For example, we wouldn't use a `shared` association from `google.co.il` to `google.com`, because `google.co.il` redirects to `accounts.google.com` for sign in.)
- [N/A] If using `from` and `to`, the new group, the `from` domain(s) redirect to the `to` domain to log in.

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -273,6 +273,12 @@
         "fromDomainsAreObsoleted": true
     },
     {
+        "shared": [
+            "nelnet.com",
+            "nelnet.net"
+        ]
+    },
+    {
         "from": [
             "nordvpn.com",
             "nordpass.com"


Sameesunkaria:
Hi @Tschrock 

Can you run `tools/convert-shared-credential-to-legacy-format.rb` to update `quirks/websites-with-shared-credential-backends.json` which is a generated file?

rmondello:
Hi @Tschrock! I'm so sorry that it's taken us so long to get back to this PR.

It looks like these websites have changed pretty significantly since you contributed this. Do you still think it makes sense for us to review and merge this? If so, we'll take another look. If not, feel free to close.

And again, sorry!

Tschrock:
Hmm, seems they rearranged their websites again.
I can't find any login on these sites anymore, so I don't think this PR is still needed.
It sounds like some accounts transfered to `auth.nelnet.studentaid.gov`, and some transfered to `sloanservicing.com`. Idk if it's a good idea to link those since the accounts have been been split apart - my account only works on the first one.

PR #703, opened 2023-07-10
Add max.com to the shared credential backends list for HBO sites

### LINKED ISSUE #701: hbonow.com and max.com are the same site
We should add a shared credential backend quirk for this.

erynofwales:
Closes #701.

### Overall Checklist

### DIFF
--- a/quirks/shared-credentials-historical.json
+++ b/quirks/shared-credentials-historical.json
@@ -283,13 +283,6 @@
     },
     {
         "shared": [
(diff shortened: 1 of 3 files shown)


rmondello (changes requested):
Needs to pass the linter. Looks like there’s already an entry for hbo.com that you need to reconcile this with.

rmondello (approved):

PR #718, opened 2023-10-31
Remove Ring from Amazon shared credentials

plttn:
Based on some research I've done I don't see any immediate evidence of Ring ever explicitly sharing with an Amazon account. While it's possible to link the accounts for features like Alexa, sources/documentation indicate explicitly that the Ring account remains just a Ring account.

Sources:

https://blog.ring.com/about-ring/ring-now-part-amazon-family/
> Nothing! Your Ring account and the Ring App will stay the same, so you’ll continue to have a superior Ring experience.

https://support.ring.com/hc/en-us/articles/360034187232-Linking-Your-Ring-and-Amazon-Accounts

CleanShot 2023-10-31 at 10 43 14@2x

(no mention of shared credentials)

Testing with my Amazon account credentials results in a "check your username and password" error.

### Overall Checklist

- [x] You believe that the domains were associated at some point in the past and can explain that relationship

### DIFF
--- a/quirks/shared-credentials-historical.json
+++ b/quirks/shared-credentials-historical.json
@@ -43,8 +43,7 @@
             "amazon.sa",
             "amazon.sg",
             "amazon.se",
-            "amazon.pl",
-            "ring.com"
+            "amazon.pl"
         ]
     },
     {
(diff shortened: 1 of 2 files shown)


Sameesunkaria:
Can you please run the `tools/convert-shared-credential-to-legacy-format.rb` script to update some generated files?

Sameesunkaria (approved):

PR #741, opened 2024-01-18
Add password rules for vince.com

dmmaslenn:
This website does not support `-` special character in password.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -920,6 +920,9 @@
     "vetsfirstchoice.com": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; allowed: [?!@$%^+=&];"
     },
+    "vince.com": {
+        "password-rules": "minlength: 8; required: digit; required: lower; required: upper; required: [$%/(){}=?!.,_*|+~#[]];"
+    },
     "virginmobile.ca": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$@];"
     },


quirks/password-rules.json
```diff
@@ -782,6 +782,9 @@
     "vetsfirstchoice.com": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; allowed: [?!@$%^+=&];"
     },
+    "vince.com": {
+        "password-rules": "minlength: 8; required: digit, lower, upper, [$%/()[]{}=?!.,_*|+~#];"
```
rmondello, line 786: What your rules says:
Required: one of a { digit, lower, upper, or special character from this set }

What the site says:

> Required: 8 Characters
> One capital and 1 lowercase letter
> At least 1 number
> At least one special character $%/()[]{}=?!.,-_*|+~#

Which translates to pseudo-code:

Min-length: 8
Required: digit
Required: lower
Required: upper
Required: one of these: $%/()[]{}=?!.,-_*|+~#

Also, `]` is treated specially in our format. It has to be the *last* character of a character class.

With all of that, I think your rule should be:

```
minlength: 8; required: digit; required: lower; required: upper; required: [$%/(){}=?!.,_*|+~#[]];
```

rmondello (changes requested):
I'd love to review this with some changes!

quirks/password-rules.json
jessieberlin: @rmondello Please do not be fooled by what the site visually claims to support. Their JS validates as follows: {length: /\S{8,}/, loweruppercase: /(?=.*[A-Z])(?=.*[a-z])/, number: /\d/, special: /[$%/()\[\]{}=?!.,_*|+~#]/}

Please do not put back in the "-" as it won't work.

rmondello (approved):

PR #752, opened 2024-02-26
Fix the CI script for linting shared-credentials.json

Sameesunkaria:
Use `jq` to sort the shared-credentials.json first by the list of "shared" domains and then the "from" domains.

### Overall Checklist


rmondello:
Doesn’t this change the meaning of the current contract?

PR #765, opened 2024-05-28
Added password rules for ana.co.jp

rohitpoks:
Added rules that conform to the requirements from the sign up page (https://www.ana.co.jp/en/us/amc/join-amc/).

Below are screenshots from the sign up page demonstrating these requirements:

Screenshot 2024-05-28 at 4 10 28 PM

Screenshot 2024-05-29 at 4 12 23 AM

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -41,6 +41,9 @@
     "americanexpress.com": {
         "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 4; required: lower, upper; required: digit; allowed: [%&_?#=];"
     },
+    "ana.co.jp": {
+        "password-rules": "minlength: 8; maxlength: 16; required: digit; required: upper,lower;"
+    },
     "anatel.gov.br": {
         "password-rules": "minlength: 6; maxlength: 15; allowed: lower, upper, digit;"
     },


quirks/password-rules.json
```diff
@@ -41,6 +41,9 @@
     "americanexpress.com": {
         "password-rules": "minlength: 8; maxlength: 20; max-consecutive: 4; required: lower, upper; required: digit; allowed: [%&_?#=];"
     },
+    "ana.co.jp": {
+        "password-rules": "minlength: 8; maxlength: 16; required: digit; required: upper,lower;"
```
aznair, line 45: What I got from the screenshot is that only alphanumeric characters are allowed, in which case shouldn't this use "allowed:" like the one below?

rohitpoks: Site requires a combination of alphabets and numbers (see image below). Attached it to the original PR comment
Screenshot 2024-05-29 at 4 12 23 AM

rohitpoks: Yeah, it requires passwords to have both a number and an alphabet. Passwords such as "abcdefghij" and "123456789" spit out the error message "Please enter your web password by combining alphabets and numbers". I believe the error message is kind of vague because it's translated from Japanese.

jmulani (approved):

PR #769, opened 2024-05-31
Add quirk for clarksoneyecare.com

vcsjones:
### Overall Checklist

#### for password-rules.json

<details>
<summary>Collapsing unrelated checklist</summary>

</details>

The patient portal for clarksoneyecare.com requires special characters, but of the permitted special characters, a hyphen (-) is not permitted. This results in Apple's default generated passwords not meeting the password requirements.

Here is a screenshot of the password rules:

Screenshot 2024-05-31 at 3 18 13 PM

Lowercase letters are required too, but since the website only shows you the rules on validation, I had to put something in the textbox to tell me what the requirements were, which I used a single letter "a" for that. That is why `required: lower` is included though the error message does not indicate it is required.

The website also (seemingly) has no maximum password length, so that was omitted.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -176,6 +176,9 @@
     "claimlookup.com": {
         "password-rules": "minlength: 8; maxlength: 16; required: lower; required: upper; required: digit; required: [@#$%^&+=!];"
     },
+    "clarksoneyecare.com": {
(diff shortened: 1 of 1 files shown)


dmmaslenn:
I've updated this PR since in the password requirements for this website it is not explicitly stated that lower case characters are required, but they are allowed.
@vcsjones please let us know if we missed something and lowercase characters need to be required.

rmondello (approved):

PR #772, opened 2024-06-23
Added shared credentials for Digi

revolter:
### Overall Checklist

#### for password-rules.json

  `digi.ro` | `digionline.ro`
  --- | ---
  https://who.is/whois/digi.ro | https://who.is/whois/digionline.ro
  - `digiromania.ro` redirects to `digi.ro`: https://digiromania.ro
  - `rcs-rds.ro` redirects to `digi.ro`: https://rcs-rds.ro


rmondello:
One of the linters is failing because we currently don’t allow overlapping rules. I’m thinking about what to do with this.

revolter:
Want some help?

quirks/shared-credentials.json
```diff
@@ -135,6 +135,22 @@
             "diversalertnetwork.org"
         ]
     },
+    {
+        "shared": [
+            "digi.ro",
+            "digionline.ro"
```
rmondello, line 141: Maybe the expedient thing to do here is move “digionline.ro” into the “to” of the other rule, even if it’s not 100% accurate. What do you think?

revolter: Makes sense

revolter: Actually, `digionline.ro` is NOT obsoleted, so I'm not sure what implications does this change have.

Would
```json
{
    "from": [
        "digionline.ro",
    ],
    "to": [
        "digi.ro"
    ],
},
{
    "from": [
        "digiromania.ro",
        "rcs-rds.ro"
    ],
    "to": [
        "digi.ro"
    ],
    "fromDomainsAreObsoleted": true
}
```
work?

revolter: Nope, it wouldn't work:
```sh
$ ruby .github/workflows/lint-scripts/websites-shared-credentials-duplicates.rb
The domain 'digi.ro' appears more than once!
```
:(

quirks/shared-credentials.json
```diff
@@ -135,6 +135,17 @@
             "diversalertnetwork.org"
         ]
     },
+    {
+        "from": [
+            "digiromania.ro",
```
rmondello, line 140: I don't see digiromania.ro redirecting. When I load it, it shows a page at that domain!

rmondello (changes requested):
Can you look into digiromania.ro not redirecting?

quirks/shared-credentials.json
revolter: Huh 🤔 Even in incognito?

```bash
$ curl --include digiromania.ro
HTTP/1.1 301 Moved Permanently
Date: Sat, 28 Sep 2024 07:24:40 GMT
Content-Type: text/html
Content-Length: 162
Connection: keep-alive
Location: https://www.digi.ro/
Server: RDS-WebServer v2

<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

rmondello:
Closing for now. Feel free to re-open if you're interested in trying to get this landed.

revolter:
I do. I was waiting for a reply from you here: https://github.com/apple/password-manager-resources/pull/772#discussion_r1779435288

rmondello (approved):

PR #773, opened 2024-06-23
Added shared credentials for Gazduire.NET

revolter:
### Overall Checklist

#### for password-rules.json

  `Tes Euro Media SRL`:
  - https://who.is/whois/gazduire.com.ro
  - https://who.is/whois/gazduire.net
  - https://who.is/whois/admin.ro
  `gazduire.com.ro` redirects to `gazduire.net`, not to `admin.ro`, but `gazduire.net` redirects to `admin.ro` when clicking the `Login` button

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -236,6 +236,16 @@
     },
     {
         "from": [
+            "gazduire.com.ro",
+            "gazduire.net"
+        ],
+        "to": [
+            "admin.ro"
+        ],
+        "fromDomainsAreObsoleted": true
+    },
+    {
+        "from": [
             "hbo.com",
             "hbomax.com",
             "hbonow.com"
--- a/quirks/websites-with-shared-credential-backends.json
+++ b/quirks/websites-with-shared-credential-backends.json
@@ -379,6 +379,11 @@
         "swarmapp.com"
     ],
     [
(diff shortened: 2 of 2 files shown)


quirks/shared-credentials.json
```diff
@@ -234,6 +234,16 @@
         ],
         "fromDomainsAreObsoleted": true
     },
+    {
+        "from": [
+            "gazduire.com.ro",
```
rmondello, line 239: I see this redirecting to gazduire.net

quirks/shared-credentials.json
```diff
@@ -234,6 +234,16 @@
...
+            "gazduire.net"
```
rmondello, line 240: And this staying on itself.

quirks/shared-credentials.json
```diff
@@ -234,6 +234,16 @@
...
+        ],
+        "to": [
+            "admin.ro"
```
rmondello, line 243: Oh, but both redirect to admin.ro when it comes time to sign in!

rmondello (approved):

quirks/shared-credentials.json
revolter: Yeah, I explained it in the description, but buried after the last checklist item 🙈

PR #788, opened 2024-07-05
yodlee-chase

jla3378:
### Overall Checklist

Unsure whether this behavior is the same across all Yodlee embeds but Chase Bank uses Yodlee as their third-party financial institution connection provider and at least with Chase it's not on Yodlee's domain but is implemented as a subdomain of chase.com (`aggregation-yodlee.chase.com`)

Screenshot 2024-07-04 at 5 33 07 PM


erynofwales:
Since Chase uses a subdomain of chase.com for this embed, it's "same site" with chase.com, and doesn't qualify for inclusion in this file.

PR #789, opened 2024-07-16
Add a password rule for pret.com

rmondello:
From the website:

> Your password must…
> have 12 or more characters
> include letters a-z
> include numbers 0-9
> have at least one of these special characters @$!%*#?&

### Overall Checklist

#### for password-rules.json

Untitled

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -779,6 +779,9 @@
     "prestocard.ca": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit,[!\"#$%&'()*+,<>?@];"
     },
+    "pret.com": {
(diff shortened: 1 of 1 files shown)


dmmaslenn (approved):

quirks/password-rules.json
```diff
@@ -779,6 +779,9 @@
     "prestocard.ca": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit,[!\"#$%&'()*+,<>?@];"
     },
+    "pret.com": {
+        "password-rules": "minlength: 12; required: lower; required: digit; required: [@$!%*#?&]; allowed: lower;"
```
dmmaslenn, line 783: Do you need both `required: lower` and `allowed: lower`?

rmondello: I don’t! I actually intended for the allowed to be for upper!

PR #798, opened 2024-08-12
Added rules for cb2.com and umterps.evenue.net

### LINKED ISSUE #795: Need password quirks for www.cb2.com and umterps.evenue.net
Our suggested passwords don't work with the following websites:

https://umterps.evenue.net/myaccount/createaccount?siteId=ev_umd&linkID=umd&continue=%2F
https://www.cb2.com/account/create-account
Screenshot 2024-08-08 at 1 01 32 PM
Screenshot 2024-08-08 at 1 01 37 PM

rohitpoks:
This PR closes #795

Requirements for cb2.com:
image

Requirements for umterps.evenue.net:
image

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -156,7 +156,7 @@
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%&*?@_];"
     },
     "cb2.com": {
-        "password-rules": "minlength: 7; maxlength: 18; required: lower, upper; required: digit;"
+        "password-rules": "minlength: 9; required: lower, upper; required: digit; required: [!#*_%.$];"
     },
     "ccs-grp.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -924,7 +924,7 @@
         "password-rules": "minlength: 12; maxlength: 30; required: lower; required: upper; required: digit; required: [!@#$%^&*()+];"
     },
     "umterps.evenue.net": {
-        "password-rules": "minlength: 4; maxlength: 12;"
+        "password-rules": "minlength: 14; required: digit; required: upper; required: lower; required: special;"
```
rmondello, line 927: Have you been able to verify that the website supports every character in the `special` set? In my experience, very few websites accept all of them.

rohitpoks: Good catch, changed the commit to only include special characters accepted by the website!

rmondello (approved):

PR #799, opened 2024-08-21
Added amazon.com to change password URL quirks

danielleflinn:
### Overall Checklist

- [X] The URL either makes the experience better or no worse than being directed to just the domain in a non-logged-in state


quirks/change-password-URLs.json
```diff
@@ -20,6 +20,7 @@
     "allianz.com.br": "https://www.allianz.com.br/alteracao-de-password-ecliente",
     "allrecipes.com": "https://www.allrecipes.com/account/profile#/change-password",
     "alternate.de": "https://www.alternate.de/html/myAccount/account/basicData.html",
+    "amazon.com": "https://www.amazon.com/ax/account/manage",
```
Sameesunkaria, line 23: Thank you @danielleflinn :)

If you'd like, this would be a good opportunity to add the equivalent urls for the rest of the global amazon domains. You can find a list in `websites-with-shared-credential-backends.json`.

Sameesunkaria (approved):

Maitha990:
Stop sending emails

2k.’

quirks/change-password-URLs.json
danielleflinn: Great idea; added!

Sameesunkaria:
Looks like the domains aren't sorted alphabetically. You can run the `./tools/autoformat-json-files.rb` script to re-format the files.

danielleflinn:
Updated! 

Some general project feedback: it seems a bit odd that `websites-with-shared-credential-backends.json ` are not sorted alphabetically while the change-urls are. My first PR I just copied and pasted the order from the other file.

Sameesunkaria:
Thank you so much for following through with this back and forth @danielleflinn 😅 

>Some general project feedback: it seems a bit odd that `websites-with-shared-credential-backends.json`  are not sorted alphabetically while the change-urls are. My first PR I just copied and pasted the order from the other file.

I somewhat agree with this. Our current ordering seems to be arbitrary, with the first domain often being the most prominent one. This allows, for example, "wikipedia.org" to be ordered with W instead of M (for "mediawiki.org"), but as you have found out this ordering is inconsistent across files. I don't have a strong opinion here, but if you feel differently, an issue would be an excellent place to start.

Sameesunkaria (approved):

rmondello (approved):

PR #802, opened 2024-08-24
Added Telegram domains to Shared Credentials

gracjankn:
### Overall Checklist

Telegram.me is used to autofill credentials in Telegram for iOS. Telegram.org is used for the web app (https://web.telegram.org).

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -445,10 +445,13 @@
         ]
     },
     {
-        "shared": [
-            "telegram.org",
+        "from": [
             "telegram.me"
-        ]
+        ],
+        "to": [
+            "telegram.org"
+        ],
+        "fromDomainsAreObsoleted": false
     },
     {
         "shared": [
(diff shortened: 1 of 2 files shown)


quirks/shared-credentials.json
```diff
@@ -446,7 +446,13 @@
     },
     {
         "shared": [
-            "ting.com", 
+            "telegram.org",
+            "telegram.me"
+        ]
```
rmondello, line 451: From what I can tell, telegram.me redirects to telegram.org, so a “from” / “to” relationship would be more correct here.

rmondello (changes requested):

quirks/shared-credentials.json
gracjankn: It does redirect when opened in the browser, however Password Autofill on iOS suggests an account for login to the Telegram app only if "telegram.me" is also added to the domains in the Passwords app.

Would "from" or "to" relationship work in this case?

rmondello: On iOS, we will interpret this correctly and do exactly what you want. :)

gracjankn: I changed the relationship type in the new commit. Does that look right?

rmondello (approved):

Issue #695, opened 2023-07-05
Heroku.com data isn’t offered on verify.salesforce.com; add a quirk between the websites?


rmondello:
Heroku.com data isn’t offered on verify.salesforce.com; add a quirk between the websites?

Real-world impact:
https://hachyderm.io/@marcomasser/110654679935725507

PR #828, opened 2024-10-07
Allow totp from Heroku to be suggested where it is validated

louim:
### Overall Checklist

Hello! This is my attempt to fix https://github.com/apple/password-manager-resources/issues/695

Heroku uses `verify.salesforce.com` for TOTP validation. If I understood the file syntax correctly, this change would allow password managers to suggest TOTP from Heroku credentials when the user is on `verify.salesforce.com`.

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -301,6 +301,14 @@
     },
     {
         "from": [
+            "heroku.com"
+        ],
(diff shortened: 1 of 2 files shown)


rmondello (approved):

rmondello:
Thanks @louim!

PR #849, opened 2024-11-07
aircanada.com & aeroplan.rewardops.com share credentials

irew:
### Overall Checklist


irew:
I can't find any evidence other than an anecdotal that you can login with the same credentials on both. Clicking to sign in on both sites offers up this UX:

image

image

PR #854, opened 2024-11-15
Add password quirk for kennedy-center.org

vcsjones:
### Overall Checklist

#### for password-rules.json

Here is a screenshot of the Kennedy Center website with its password rules.

Screenshot 2024-11-15 at 11 40 44 AM

**Their explanation text is incomplete.**

The site also requires a lower-case letter, I think it is just implied. Their actual password validation regex is `^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*?&#])[A-Za-z\d@$!%*?&#]{8,}$`. You can see that one of the criteria is `[a-z]`.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -626,6 +626,9 @@
     "keldoc.com": {
         "password-rules": "minlength: 12; required: lower; required: upper; required: digit; required: [!@#$%^&*];"
     },
+    "kennedy-center.org": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%&*?@];"
+    },
     "key.harvard.edu": {
         "password-rules": "minlength: 10; maxlength: 100; required: lower; required: upper; required: digit; allowed: [-@_#!&$`%*+()./,;~:{}|?>=<^[']];"
     },


quirks/password-rules.json
```diff
@@ -626,6 +626,9 @@
     "keldoc.com": {
         "password-rules": "minlength: 12; required: lower; required: upper; required: digit; required: [!@#$%^&*];"
     },
+    "kennedy-center.org": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%&*?@];"
```
rmondello, line 630: Killer commit message, @vcsjones! Good enough that it makes it really easy to give you the only feedback I have: this rule should have an "allowed: lower;" at the end.

rmondello (changes requested):

quirks/password-rules.json
vcsjones: @rmondello a lower is also required (above where I said "The site also requires a lower-case letter, I think it is just implied.").

Should the rule have both a `required: lower` _and_ an `allowed: lower` at the end?

rmondello: Wow, I failed to read this properly! This is absolutely perfect as-is!

rmondello (approved):

PR #865, opened 2025-01-06
Add password rules and password change page for Mountain Warehouse

homsar:
### Overall Checklist

#### for password-rules.json

- [X] The URL either makes the experience better or no worse than being directed to just the domain in a non-logged-in state

The default password rules generate passwords that are rejected by mountainwarehouse.com as too long; the quirk added by this PR restricts the length such that generated passwords are accepted.

screenshot

### DIFF
--- a/quirks/change-password-URLs.json
+++ b/quirks/change-password-URLs.json
@@ -269,6 +269,7 @@
     "michaels.com": "https://www.michaels.com/on/demandware.store/Sites-MichaelsUS-Site/default/Account-EditProfile",
     "microsoft.com": "https://account.live.com/password/Change",
     "mlb.com": "https://www.mlb.com/account/general",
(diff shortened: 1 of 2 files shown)


rmondello:
This change is a clear improvement. But they don't allow any special characters?

homsar:
Thanks for pointing that out—I’d forgotten about the `allowed` option.

Experimenting with allowed symbols also gave me a slightly altered version of the password rules with constraints that I don’t think can be expressed in the quirk syntax—the quirk as written encodes the requirement in the screenshot above plus allowing most* of the allowed symbols from the below, which is sufficient to satisfy either set of rules, but is technically over constrained. 

\* I excluded square brackets as those appear to be reserved as sentinel values so can’t be included in a character class.

screenshot

jmulani (approved):

PR #880, opened 2025-03-27
Add GameFAQs

GHVirus:
### Overall Checklist

#### for password-rules.json

----
IMG_7792
IMG_7793

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -380,9 +380,6 @@
     "fuelrewards.com": {
         "password-rules": "minlength: 8; maxlength: 16; allowed: upper,lower,digit,[!#$%@];"
     },
-    "gamefaqs.gamespot.com": {
-        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; allowed: ascii-printable;"
-    },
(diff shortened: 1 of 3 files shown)


Avilanator3 (approved):

quirks/password-rules.json
```diff
@@ -380,6 +380,9 @@
     "fuelrewards.com": {
         "password-rules": "minlength: 8; maxlength: 16; allowed: upper,lower,digit,[!#$%@];"
     },
+    "gamefaqs.gamespot.com": {
+        "password-rules": "minlength: 8; required: lower; required: upper; required: digit; allowed: ascii-printable;"
+    },
```
taylortrimble, line 385: I think this rule qualifies as "particularly standard and obvious for password managers." Contributions to password rules should be limited to services [which don't accept a password generated by your password manager with its default settings or reasonable configuration](https://github.com/apple/password-manager-resources/blob/main/CONTRIBUTING.md#contributing-password-rules).

rmondello: Agreed. We should remove this.

rmondello (approved):

PR #885, opened 2025-04-18
Add Zumper/PadMapper to shared credentials

dloehr:
### Overall Checklist

##### zumperrentals.com

Clicking log in on zumperrentals.com takes the user to zumper.com to log in:

https://github.com/user-attachments/assets/bc56d398-d940-4049-b8cb-8e6caf977841

##### zumper.com and padmapper.com

Evidence that zumper.com and padmapper.com share the same credential backend includes:

1. Attempting to create an account on padmapper.com using an email address that already has an account gives the error message, "A PadMapper/Zumper user with this email already exists"
    image
2. Signing in on either site redirects through all 3 sites, so that the user will be signed into the same account on all of them:
    image

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -469,6 +469,13 @@
         ]
     },
     {
+        "shared": [
+            "padmapper.com",
+            "zumper.com",
+            "zumperrentals.com"
+        ]
+    },
+    {
         "from": [
             "parkmobile.us"
         ],


rmondello (approved):

rmondello:
@dloehr This change looks sensible to me, but one of our checks failed. Can you take a look and see if you want to make a change? (Let me know if something doesn’t make sense.)

> Run ruby .github/workflows/lint-scripts/websites-shared-credentials-duplicates.rb
> The domain 'zumper.com' appears more than once!

dloehr:
Thanks @rmondello. I can think of 3 possibilities, do any of them seem acceptable/preferable to you?

1. Combine the 2 rules I'm adding into just
    ```json
    {
        "shared": [
            "padmapper.com",
            "zumper.com",
            "zumperrentals.com"
        ]
    },
    ```
    That seems slightly incorrect to me, since zumperrentals.com redirects to zumper.com for authentication, and users no longer enter their credentials on zumperrentals.com
2. Remove the zumperrentals.com rule altogether. zumperrentals.com only had its own sign in form (which shared a backend with zumper.com and padmapper.com) for about 4 months, so I think it's relatively unimportant to mention it here
3. Adjust the lint rule from, "A domain may only be mentioned once across `to`, `from`, and `shared`" to "A domain may be mentioned either (1) in `from` only once or (2) up to once in each of `shared` and `to`"

taylortrimble:
It sounds like for a period of 4 months, users may have created a password in their password manager for zumperrentals.com that's valid now on zumper.com and padmapper.com. So I think you can have your rule with the 3 shared (your option 1).

rmondello (approved):

rmondello:
Thank you @dloehr!

PR #889, opened 2025-04-25
Add threads.com

conmas:
Threads has updated their domain from threads.net to threads.com, adding this to shared credentials. 

Source: https://about.fb.com/news/2025/04/new-features-threads-web-experience/

### Overall Checklist

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -342,7 +342,8 @@
     {
         "shared": [
             "instagram.com",
-            "threads.net"
+            "threads.net",
+            "threads.com"
         ]
     },
     {


conmas:
Should there be a `from / to` association between `threads.net` and `threads.com`, and then separately a `shared` association between `instagram.com` and `threads.com`? Not sure how to handle this scenario, there weren't other examples in the list.

aabosh (approved):

rmondello:
> Should there be a from / to association between threads.net and threads.com, and then separately a shared association between instagram.com and threads.com? Not sure how to handle this scenario, there weren't other examples in the list.

This is a weakness in our schema. I don’t want to block this change on this, so let’s go with the less granular “shared” for now.

rmondello:
Thanks @conmas!

PR #899, opened 2025-05-03
Add safeway.com password quirks

taylortrimble:
From the password reset page:

- Between 8–40 characters.
- At least one upper case letter.
- At least one symbol (!@#$%^&*).

Tested with all provided symbols, including at the front of the password, and with passwords not containing any digits or lower case letters.

Screenshot:
Screenshot 2025-05-02 at 11 38 57 PM

### Overall Checklist

#### for password-rules.json


taylortrimble:
Note: because of safeway.com rules, default passwords generated by Safari like:

* `tyckyp-4bUhxe-jymroz`
* `nahgeN-4huwdi-gadjos`
* `wumky6-hippef-kubHux`

are not allowed because hyphens aren't allowed.

rmondello (approved):

rmondello:
We should consider mirroring these rules to all of the other Albertsons websites.

PR #946, opened 2025-09-01
Update password-rules.json for BlueCross BlueShield of NM

new-mexico-worm-farmer:
### Overall Checklist

#### for password-rules.json

BlueCross BlueShield of New Mexico has password rules that break common assumptions (in particular, you can't use a dash/hyphen, which breaks the usual sorts of secure passwords generated on Apple platforms)

There are additional rules they enforce that I could not figure out how to encode in these rules (like banning 5 or more digits and some specific phrases).  However, using the Password Rules Validation Tool it seems like these are unlike to be generated.

bcbsnm-pw-rules


rmondello:
Thanks for the contribution.

rmondello (approved):

Issue #921, opened 2025-06-22
Password for Westfield.com


mpwg:
The suggested password lacks a "special character" 
I can't find the requirements again, was only on the forgot password page but I think it was max 20 characters,  special characters, and the usual (upper/lower, number and I think at least 8 chars)

PR #949, opened 2025-09-02
Update password-rules.json for id.westfield.com

new-mexico-worm-farmer:
### Overall Checklist

#### for password-rules.json

per the request in #921 

This website will recognize _many_ special characters, but not the `-` character!  As you can see in the javascript they're using, technically there are other characters which could be used.  However, the Password Rules Validation Tool did not seem to appreciate many of them (such as '«'), so this only includes the special characters that the validation tool filtered down to. 

westfield-pw-reqs

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -623,6 +623,9 @@
     "id.sonyentertainmentnetwork.com": {
         "password-rules": "minlength: 8; maxlength: 30; required: lower, upper; required: digit; allowed: [-!@#^&*=+;:];"
     },
+    "id.westfield.com": {
+        "password-rules": "minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; required: [!\"#&'()*,./:;?@[\\^_`{|}~];"
+    },
     "identity.codesignal.com": {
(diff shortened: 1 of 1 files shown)


quirks/password-rules.json
```diff
@@ -623,6 +623,9 @@
     "id.sonyentertainmentnetwork.com": {
         "password-rules": "minlength: 8; maxlength: 30; required: lower, upper; required: digit; allowed: [-!@#^&*=+;:];"
     },
+    "id.westfield.com": {
+        "password-rules": "minlength: 9; maxlength: 20; required: lower; required: upper; required: digit; required: [!\"#&'()*,./:;?@[\\^_`{|}~];"
```
rmondello, line 627: There are two very special quirks of character classes (that I really need to add a linter to detect).

If `-` is present, it must be the first character in the class, and if `]` is present, it must be the last character.

Example:
`[- !]]`

rmondello: BUT it looks like `]` isn’t included in the code from the website, soooooo, this is good!

rmondello (approved):

rmondello:
Keep ‘em coming, @new-mexico-worm-farmer! I’m having fun.

PR #952, opened 2025-09-07
Update password-rules.json for zenni optical

new-mexico-worm-farmer:
### Overall Checklist

#### for password-rules.json

This website only requires upper/lower/digit: 
zenni-pw-rules

However, their client-side javascript appears to allow essentially any special character:

```javascript
        class FusionAuthPasswordChecker {
          constructor(minLength, maxLength, requireMixedCase, requireNonAlpha, requireNumber) {
            this.rules = { minLength, maxLength, requireMixedCase, requireNonAlpha, requireNumber };
          }

          validate(password) {
            const messages = [];
            let valid = true;

            if (password.length < this.rules.minLength || password.length > this.rules.maxLength) {
              messages.push(this.rules.minLength + ' to ' + this.rules.maxLength + ' characters long');
              valid = false;
            }

            if (this.rules.requireMixedCase && !/[A-Z]/.test(password)) {
              messages.push('One uppercase character');
              valid = false;
            }

            if (this.rules.requireMixedCase && !/[a-z]/.test(password)) {
              messages.push('One lowercase character');
              valid = false;
            }

            if (this.rules.requireNumber && !/[0-9]/.test(password)) {
              messages.push('One number');
              valid = false;
            }

            if (this.rules.requireNonAlpha && !/[^A-Za-z0-9]/.test(password)) {
              messages.push('One special character.');
              valid = false;
            }

            return { valid, messages };
          }
        }

        const minLength = 8;
        const maxLength = 14; 
        const requireMixedCase = true;
        const requireNonAlpha = false;
        const requireNumber = true;
```

For example:
howdy-pw

I tested a dozen passwords from the Password validation tool and the client-side javascript accepted them all.  However, I did not want to potentially mess up my account if this validation isn't great, so I did not submit any changes for any server-side validation, nor did I want to create a dummy account.

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -113,6 +113,9 @@
     "auth.readymag.com": {
         "password-rules": "minlength: 8; maxlength: 128; required: lower; required: upper; allowed: special;"
     },
+    "auth.zennioptical.com": {
+        "password-rules": "minlength: 8; maxlength: 14; required: lower; required: upper; required: digit; allowed: special;"
+    },
     "autify.com": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!\"#$%&'()*+,./:;<=>?@[^_`{|}~]];"
     },


new-mexico-worm-farmer:
I should add that the issue to be addressed here is that because the password is relatively short, the default behavior sometimes (but not always) generates a password that is missing a required character class (e.g. number).

PR #953, opened 2025-09-08
Update password-rules.json for NM Lottery website

new-mexico-worm-farmer:
### Overall Checklist

#### for password-rules.json

The rules given on the website:
nmlottery-rules

This matches the client-side validation javascript:
nmlottery-javascript

Experimentally, I verified that they allow passwords with special characters.  Technically, only one of upper or lower would be required, but it seems better to have them both as required (along with the optional special character) vs. trying to strictly encode rules that would only serve to generate weaker passwords (but willing to make changes if that's not in the spirit).


rmondello:
@new-mexico-worm-farmer Thanks again!

PR #1000, opened 2026-01-07
Add support for Ecosia Browser

ecotopian:
### Overall Checklist

Ecosia Browser is a privacy focused browser which puts 100% of its profits into climate action -- https://www.ecosia.org/browser

### DIFF
--- a/quirks/web-browser-extension-distribution-information.json
+++ b/quirks/web-browser-extension-distribution-information.json
@@ -1116,6 +1116,25 @@
             "supported_store_identifiers": [
                 3
             ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


rmondello:
@ecotopian Thank you for this contribution!

If you haven't already, you might want to check this out, too: https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.developer.web-browser.public-key-credential

ecotopian:
Thank you loads for the super quick action @rmondello 💚 The planet will thank you 🌳 

Can one/you already say when the change would be in effect? I understood that for MacOS a system update is needed. How is it for Windows?

PR #1091, opened 2026-05-11
Add change password URL and password rule quirks for vagaro.com, weve…

dmmaslenn:
…rse.io

Screenshot 2026-05-10 at 12 14 33 AM

### Overall Checklist

#### for password-rules.json


quirks/password-rules.json
```diff
@@ -1154,6 +1154,9 @@
     "usps.com": {
         "password-rules": "minlength: 8; maxlength: 50; max-consecutive: 2; required: lower; required: upper; required: digit; allowed: [-!\"#&'()+,./?@];"
     },
+    "vagaro.com": {
+        "password-rules": "minlength: 9; required: lower; required: upper; required: digit; required: [@$!%*?&]; allowed: lower, upper, digit, [@$!%*?&];"
```
aabosh: Do we need the `allowed` here?

aabosh (changes requested):

quirks/change-password-URLs.json
```diff
@@ -575,6 +576,7 @@
     "weather.com": "https://weather.com/member/settings",
     "webmd.com": "https://member.webmd.com/password-reset",
     "welt.de": "https://signin.auth.welt.de/change-password",
+    "weverse.io": "https://weverse.io/more",
```
rmondello: Is this a high quality change password URL?

quirks/password-rules.json
rmondello: We do not. We should remove it.

rmondello (changes requested):

quirks/password-rules.json
dmmaslenn: Removed.

quirks/change-password-URLs.json
dmmaslenn: Added more specific URL.

rmondello (approved):

aabosh (approved):

PR #1093, opened 2026-05-12
Add change password URLs

dmmaslenn:
for cookpad.com, emirates.com, zalando.de, zara.com

### Overall Checklist


whsieh:
https://www.zalando.de/myaccount/change-password/ seems to navigate to an error page for me. As for the other URLs, they all seem locale-specific (English / US)

quirks/change-password-URLs.json
```diff
@@ -9,6 +9,7 @@
     "abc.es": "https://areapersonal.abc.es/gestion/cambiar-contrasena.html",
     "abc.net.au": "https://account.abc.net.au/login-details",
     "abcmouse.com": "https://www.abcmouse.com/abc/parent-section",
+    "abercrombie.com": "https://www.abercrombie.com/shop/us",
```
aabosh: This seems to just redirect to the homepage?

aabosh (changes requested):

quirks/change-password-URLs.json
dmmaslenn: URL indeed is not right, pushed by mistake. Updated PR.

dmmaslenn:
> https://www.zalando.de/myaccount/change-password/ seems to navigate to an error page for me. As for the other URLs, they all seem locale-specific (English / US)

I have fixed URL for zalando, nice catch. URLs are locale specific since there is no generic URL that would work for every locale on these websites, if we want to cover more locales we should add additional entry per each locale.

Issue #907, opened 2025-05-08
Improve password-rules.json error catching in CI


erynofwales:
I landed a [change](https://github.com/apple/password-manager-resources/commit/e976f36e523e90a5d291ca3c6cafd2926d107f17) recently to fix a small formatting error in the password-rules.json file. This caused an error in some of our scripts:

```
Parsing password rules for hetzner.com
minlength: 8; required: lower; required: upper; required: digit, required: [-^!$%/()=?+#.,;:~*@{}_&[]];
_passwordGenerationParametersDictionary@(execjs):959:41: TypeError: null is not an object (evaluating 'requirements["PasswordMinLength"]') (ExecJS::ProgramError)
```

PR #1094, opened 2026-05-18
Improve password rules validation errors

LeSingh1:
## Summary

Adds a CI lint step that parses every `"password-rules"` value in `quirks/password-rules.json` using the existing `tools/PasswordRulesParser.js`. Any parse error is reported with the offending domain and the parser's error message, causing the `password-rules` lint job to fail.

This addresses [#907](https://github.com/apple/password-manager-resources/issues/907), where a small formatting error (a comma instead of a semicolon between two `required:` properties for `hetzner.com`) landed without CI catching it and only surfaced later as a downstream parser exception.

The new check piggybacks on the existing `password-rules` job (sort-order check) and adds two steps: `actions/setup-node@v6` and `node .github/workflows/lint-scripts/password-rules-parse.js`. No other lint behavior changes.

## How it works

`password-rules-parse.js` loads `tools/PasswordRulesParser.js` via `node:vm` into a sandbox whose `console.error` is captured. It iterates every domain in `password-rules.json`, calls `parsePasswordRules()`, and exits non-zero if any rule produced a parser error. The parser is loaded unchanged — it remains usable by Safari/WebKit consumers that expect a plain script.

## Test plan

- [x] Run `node .github/workflows/lint-scripts/password-rules-parse.js` on `main` → `OK: 419 password-rules entries parsed cleanly.` (exit 0)
- [x] Reintroduce the original `hetzner.com` comma typo from #907 → script fails with `1 domain(s) in password-rules.json failed to parse.` and identifies `hetzner.com` (exit 1)
- [x] Restore correct value → clean again (exit 0)
- [x] `python -c "import yaml; yaml.safe_load(open('.github/workflows/lint.yml'))"` → YAML valid

## Risks / follow-ups

- The script only fails on `console.error` from the parser; recoverable warnings (e.g. the `console.warn` about a `-` inside a character class) are intentionally ignored to avoid noisy false positives. If desired, a follow-up could promote those to errors.
- This does not introduce any runtime dependencies — it uses only Node built-ins (`fs`, `path`, `vm`).

### DIFF
--- a/.github/workflows/lint-scripts/password-rules-parse.js
+++ b/.github/workflows/lint-scripts/password-rules-parse.js
@@ -1,0 +1,56 @@
+// Copyright (c) 2026 Apple Inc. Licensed under MIT License.
+//
+// Validates every "password-rules" string in quirks/password-rules.json by
+// loading tools/PasswordRulesParser.js and calling parsePasswordRules() on it.
+// Any parse error (e.g. ",required" instead of "; required", as in #907) causes
+// CI to fail with a clear message identifying the offending domain.
+
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+const vm = require("vm");
+
+const repoRoot = path.resolve(__dirname, "..", "..", "..");
+const parserPath = path.join(repoRoot, "tools", "PasswordRulesParser.js");
+const rulesPath = path.join(repoRoot, "quirks", "password-rules.json");
+
+const parserSource = fs.readFileSync(parserPath, "utf8");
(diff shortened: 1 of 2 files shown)


rmondello (approved):

erynofwales (approved):

rmondello:
@LeSingh1 Awesome contribution! Thank you!

I think the next step here is to ensure that we "fail" in the CI if `-` is in a character class, but isn't the first character of the class and/or if `]` is in a character class, but isn't the last character of the class. That's a mistake that people make a lot.

PR #1095, opened 2026-05-18
Teach the new lint script, password-rules-parse.js, to fail on common…

rmondello:
… character class errors

Two existing warnings are promoted to errors and one new error is added to the parser.

### Overall Checklist

### DIFF
--- a/tools/PasswordRulesParser.js
+++ b/tools/PasswordRulesParser.js
@@ -333,8 +333,7 @@ function _parseCustomCharacterClass(input, position)
         }
 
         if (c === "-" && (position - initialPosition) > 0) {
-            // FIXME: Should this be an error?
-            console.warn("Ignoring '-'; a '-' may only appear as the first character in a character class");
+            console.error("Ignoring '-'; a '-' may only appear as the first character in a character class");
             ++position;
             continue;
         }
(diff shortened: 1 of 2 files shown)


tools/PasswordRulesParser.js
```diff
@@ -349,6 +348,12 @@ function _parseCustomCharacterClass(input, position)
     if (position < length && input[position] !== CHARACTER_CLASS_END_SENTINEL || position == length && input[position - 1] == CHARACTER_CLASS_END_SENTINEL) {
         // Fix up result; we over consumed.
         result.pop();
+        if (position < length) {
+            let nextCharacter = input[position];
+            if (!_isASCIIWhitespace(nextCharacter) && nextCharacter !== PROPERTY_VALUE_SEPARATOR && nextCharacter !== PROPERTY_SEPARATOR) {
+                console.error("A ']' inside a character class must be the last character of the class; if a literal ']' is intended, write ']]' at the end. Found unexpected '" + nextCharacter + "' after closing ']'.");
```
erynofwales: Nit: I would make this a backtick quoted string and use the interpolation syntax.
```suggestion
                console.error(`A ']' inside a character class must be the last character of the class; if a literal ']' is intended, write ']]' at the end. Found unexpected '${nextCharacter}' after closing ']'.`);
```

erynofwales (approved):

tools/PasswordRulesParser.js
rmondello: Good call.

PR #1098, opened 2026-05-26
Add NPO domains

WouterVermeijNPO:
### Overall Checklist

### DIFF
--- a/quirks/shared-credentials.json
+++ b/quirks/shared-credentials.json
@@ -877,5 +877,16 @@
             "ynab.com"
         ],
         "fromDomainsAreObsoleted": true
+    },
+        {
+        "shared": [
+            "eo.nl",
+            "id.npo.nl",
+            "kro-ncrv.nl",
+            "npo.nl",
+            "npo.nl/luister",
+            "npo.nl/start",
+            "wnl.tv"
+        ]
     }
 ]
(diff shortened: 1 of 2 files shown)


iDylanK (approved):

quirks/shared-credentials.json
```diff
@@ -355,6 +355,30 @@
...
         ]
     },
+    {
+        "from": [
+            "eo.nl",
+            "gld.nl",
+            "kro-ncrv.nl",
+            "l1.nl",
+            "npo.nl",
+            "npo.nl/start",
```
rmondello: these are supposed to be domains, so /start shouldn't be here

rmondello (changes requested):

quirks/shared-credentials.json
WouterVermeijNPO: Updated @rmondello

sander-bol (approved):
As developer working on some of these sites that leverage NPO ID, I'm happy to randomly come across this PR! 😅

@WouterVermeijNPO If there is metadata we can add to the regional  broadcaster's sites instead of relying on a quirks definition in this repo, please let me know. We're happy to facilitate better options if they are available!

rmondello (approved):

PR #1112, opened 2026-06-10
Add unit tests for PasswordRulesParser.js

Muneerali199:
## Summary

Adds 65 unit tests for `tools/PasswordRulesParser.js`, closing #98.

The test cases were adapted from the maintainer's Ruby test suite (shared in [issue #98](https://github.com/apple/password-manager-resources/issues/98#issuecomment-640653119)) and expanded with additional edge cases covering custom character classes, canonicalization, unicode handling, and invalid inputs.

## Changes

- **`tools/PasswordRulesParser.test.js`**: 65 unit tests covering:
  - Empty/whitespace-only input
  - All named character classes (upper, lower, digit, special, ascii-printable, unicode)
  - Multiple required/allowed rules with merging
  - max-consecutive (lowest wins), minlength (highest wins), maxlength (lowest wins)
  - Custom character classes with dash (`-`) and right-bracket (`]`) edge cases
  - Non-ASCII character dropping in custom classes
  - Canonicalization (full alphabet → named class)
  - Invalid/malformed inputs that should produce errors
  - Rule ordering after parsing
  - Whitespace tolerance around values

- **`.github/workflows/lint.yml`**: New `parser-unit-tests` CI job that runs the test suite on every push/PR

## Verification

```
65 of 65 tests passed.
```


tools/PasswordRulesParser.test.js
```diff
@@ -0,0 +1,290 @@
...
+        threw = true;
+    }
+    if (threw) {
+        passed++;
+    } else {
+        failed++;
+        console.error(`FAIL: ${name} - expected error but parsed successfully`);
+        console.error(`  Input: "${input}"`);
+    }
+}
```
rmondello, line 83: I think this helper conflates logging and error with rejection. But in some cases we log errors that aren't fatal and return a meaningful rule. What do you think?

tools/PasswordRulesParser.test.js
```diff
@@ -0,0 +1,290 @@
...
+// =============================================================================
+test("allowed:upper", "allowed:upper",
+    '[{"_name":"allowed","value":[{"_name":"upper"}]}]');
+test("allowed with leading whitespace", "allowed:     upper",
+    '[{"_name":"allowed","value":[{"_name":"upper"}]}]');
+
+// =============================================================================
+// Mixed required and allowed
+// =============================================================================
+test("required upper with redundant ASCII printable", "required: upper, [AZ];",
```
rmondello, line 118: I think you should change "AZ" to "ABC" so the redundancy is what sticks out, not A...Z, which someone can misread as A-Z.

rmondello (changes requested):

rmondello:
Your commit message is missing the checkbox from the normal PR template that you agree to the Developer Certificate of Origin.

PR #1119, opened 2026-06-16
Add new change password URLs

dmmaslenn:
Added change password URLs for talbots.com, starbucks.com, marca.com, box.com and fixed URL for shutterfly.com.

### Overall Checklist


quirks/change-password-URLs.json
```diff
@@ -326,6 +327,7 @@
     "lowes.com": "https://www.lowes.com/security/settings",
     "lu.ma": "https://lu.ma/settings",
     "macys.com": "https://www.macys.com/account/profile?cm_sp=macys_account-_-my_account-_-my_profile&linklocation=leftrail",
+    "marca.com": "https://www.marca.com/micuenta/v1/es/user-area/account",
```
aabosh: We shouldn't include locales in these URLs. Does `https://www.marca.com/micuenta/v1/user-area/account` work?

quirks/change-password-URLs.json
```diff
@@ -467,7 +469,7 @@
     "shoop.de": "https://www.shoop.de/einstellungen/benutzerdaten",
     "shopback.co.kr": "https://www.shopback.co.kr/account/change-password",
     "shopify.com": "https://accounts.shopify.com/lookup",
-    "shutterfly.com": "https://www.shutterfly.com/account-settings/",
+    "shutterfly.com": "https://accounts.shutterfly.com/account-settings",
```
aabosh, line 472: Is there any difference between this and the old one? They both redirect me to the same place.

aabosh (changes requested):

quirks/change-password-URLs.json
dmmaslenn: Old URL redirects to the new one, so we would skip one navigation hop.

quirks/change-password-URLs.json
dmmaslenn: Updated URL to be generic.

quirks/change-password-URLs.json
rmondello: Sounds good.

aabosh (approved):

PR #1122, opened 2026-06-18
Add change-password URLs

dmmaslenn:
Add change-password URLs for adventhealth.com, idealista.com, kahoot.it, me-qr.com, onelink.to, signupgenius.com. Fixed kohls.com - new URL lands directly on change password form, old URL opened Account Settings - Address page.

### Overall Checklist


quirks/change-password-URLs.json
```diff
@@ -473,6 +478,7 @@
     "shopify.com": "https://accounts.shopify.com/lookup",
     "shutterfly.com": "https://accounts.shutterfly.com/account-settings",
     "shutterstock.com": "https://accounts.shutterstock.com/credentials/change",
+    "signupgenius.com": "https://www.signupgenius.com/index.cfm?go=c.ResetPW",
```
rmondello, line 481: is this one actually a change password page, or does it trigger reset?

rmondello (approved):

quirks/change-password-URLs.json
dmmaslenn: Yes, it indeed loads to a page with change password form, agree that ResetPW part is misleading

PR #1123, opened 2026-06-21
Add password rules for roblox.com

gords2:
Adds password rules for roblox.com
The signup page states: "Passwords must be between 8 and 200 characters long."
Tested with lowercase only (aaaaaaaa), numbers only, special characters — all accepted.
No character type restrictions enforced beyond length.

### Overall Checklist
- [x] I agree to the project's [[Developer Certificate of Origin](https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md)](https://github.com/apple/password-manager-resources/blob/main/DEVELOPER_CERTIFICATE_OF_ORIGIN.md)
- [x] There are no [[open pull requests](https://github.com/apple/password-manager-resources/pulls)](https://github.com/apple/password-manager-resources/pulls) for the same update
#### for password-rules.json
- [x] Generated passwords have been tested from this rule using the [[Password Rules Validation Tool](https://developer.apple.com/password-rules/)](https://developer.apple.com/password-rules/)

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -986,6 +986,9 @@
     "robinhood.com": {
         "password-rules": "minlength: 10;"
     },
+    "roblox.com": {
+    "password-rules": "minlength: 8; maxlength: 200;"
+    },
     "rogers.com": {
         "password-rules": "minlength: 8; required: lower, upper; required: digit; required: [!@#$];"
     },


quirks/password-rules.json
```diff
@@ -986,6 +986,9 @@
     "robinhood.com": {
         "password-rules": "minlength: 10;"
     },
+    "roblox.com": {
+    "password-rules": "minlength: 8; maxlength: 200;"
```
rmondello, line 990: These requirements are entirely reasonable and any password manager would, by default, generate passwords that are met by them. I'm not sure they're worth incorporating into the project.

gords2: Thanks for the quick review, I'll close this one. 
The 200 character max seemed worth noting but I see your point 
that any default password manager would already handle this fine.

PR #1138, opened 2026-07-05
Add password rules for hdfc.bank.in

shamsheeryousaf:
#### Summary
- Adds a `password-rules` entry for `hdfc.bank.in`: minlength 8, maxlength 15, requires a digit and both upper/lower case, allows a specific set of special characters: @ & : | . ^ ~ # _ $ ! ) ?

#### Screenshot
Screenshot 2026-07-05 at 10 15 36 AM

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -479,6 +479,9 @@
     "hawaiianairlines.com": {
         "password-rules": "maxlength: 16;"
     },
+    "hdfc.bank.in": {
+        "password-rules": "minlength: 8; maxlength: 15; required: digit; required: upper, lower; allowed: [@&:|.^~#_$!)?];"
+    },
     "hertz-japan.com": {
         "password-rules": "minlength: 8; maxlength: 30; max-consecutive: 3; required: lower; required: upper; required: digit; required: [#$%^&!@];"
     },


rmondello:
@shamsheeryousaf Were you able to confirm that passwords generated with these rules are accepted by the website? I wanna be really sure that the website allows those special characters but doesn't require them.

shamsheeryousaf:
@rmondello Yes, I have tested with the Password Rules Validation Tool. The website allows those special characters but doesn’t require them. Sharing some screenshots below with passwords generated with the validation tool (except for the last one).

### Without special characters
image
image
image

### With allowed special characters
image

### With special characters not allowed (This should fail)
image

rmondello (approved):

quirks/password-rules.json
```diff
@@ -479,6 +479,9 @@
     "hawaiianairlines.com": {
         "password-rules": "maxlength: 16;"
     },
+    "hdfc.bank.in": {
+        "password-rules": "minlength: 8; maxlength: 15; required: digit; required: upper, lower; allowed: upper, lower, digit, [@&:|.^~#_$!)?];"
```
rmondello: All required characters are implicitly allowed, so this can be simplified to:

```
minlength: 8; maxlength: 15; required: digit; required: upper, lower; allowed: [@&:|.^~#_$!)?];
```

rmondello (changes requested):

quirks/password-rules.json
shamsheeryousaf: Ah ok! Thank you for pointing that out.

shamsheeryousaf: Amended commit with the simplified rule.

rmondello (approved):

rmondello:
@shamsheeryousaf Thanks for iterating on this!

PR #1153, opened 2026-07-18
Add change-password URLs

dmmaslenn:
Add change-password URL for bing.com, itv.com, kroger.com, microsoft.com, dafont.com, expedia.com, zendesk.com, kakao.com.

Additionally marked rentcafe.com and securecafe.com as shared credentials and added bing.com to live.com shared credentials. Added from/to relation for kakao.com -> daum.net.

### Overall Checklist


quirks/change-password-URLs.json
```diff
@@ -291,7 +293,7 @@
     "idealo.de": "https://account.idealo.de/settings",
     "ign.com": "https://www.ign.com/account/security",
     "iherb.com": "https://checkout.iherb.com/users/account-info/edit-password",
-    "ihg.com": "https://www.ihg.com/rewardsclub/gb/en/account-mgmt/personalInformation",
+    "ihg.com": "https://www.ihg.com/rewardsclub/us/en/account-mgmt/settings/login-security",
```
whsieh: I suppose this isn't a new issue, but this seems like it would require a different quirk.

quirks/change-password-URLs.json
```diff
@@ -572,6 +576,7 @@
     "ulta.com": "https://www.ulta.com/account/manage",
     "uml.edu": "https://mypassword.uml.edu/#Change",
     "umsystem.edu": "https://password.umsystem.edu/reset/",
+    "uniqlo.com": "https://www.uniqlo.com/us/en/account/password/edit",
```
whsieh: Same here — is uniqlo.com always just `us/en/`?

ahmed9959111-eng: > Same here — is uniqlo.com always just `us/en/`?

quirks/change-password-URLs.json
dmmaslenn: Good point, I have removed both uniqlo.com and ihg.com change.

ahmed9959111-eng (approved):
IMEI 354833969981618

whsieh (approved):

PR #1162, opened 2026-07-20
Reject unsupported partial ranges in regexp conversion

### LINKED ISSUE #1161: Partial alphanumeric ranges produce literal-hyphen Password Rules
Partial alphanumeric ranges can be silently converted into rules that require a literal hyphen.

## Reproduction

Verified against `91312a9`.

Required-character path:

- Pattern: `^(?=.*[a-y])[a-zA-Z0-9]{8,}$`
- Expected: `null` because the partial range is unsupported.
- Actual: `required: [-]; allowed: lower, upper, digit; minlength: 8;`

Allowed-character path:

- Pattern: `^[a-z2-9]{8,}$`
- Expected: `null` because the partial range is unsupported.
- Actual: `allowed: lower, [-]; minlength: 8;`

The control pattern `^[a-zxA-Z-y]{8,}$` currently converts accurately to `allowed: lower, upper, [-]; minlength: 8;` and should remain supported.

Rejecting the unsupported ranges avoids silently broadening their meaning or introducing a literal-hyphen requirement.

## Actual screenshot

Terminal reproduction showing two partial ranges converted into literal-hyphen rules and one valid control conversion at revision 91312a9

Kevinjohn:
Partial alphanumeric ranges can be silently converted into rules that require a literal hyphen.

### Overall Checklist

- [x] The top-level JSON objects are sorted alphabetically (N/A — no JSON changes)

Fixes #1161. This change is independent of #1160.

## Reproduction

Required-character path:

- Pattern: `^(?=.*[a-y])[a-zA-Z0-9]{8,}$`
- Expected: `null` because the partial range is unsupported.
- Actual: `required: [-]; allowed: lower, upper, digit; minlength: 8;`

Allowed-character path:

- Pattern: `^[a-z2-9]{8,}$`
- Expected: `null` because the partial range is unsupported.
- Actual: `allowed: lower, [-]; minlength: 8;`

## Change

- Validate alphanumeric ranges before converting character classes: any `X-Y` range with alphanumeric endpoints must be exactly `a-z`, `A-Z`, or `0-9`.
- Reject unsupported partial ranges in both lookaheads and the main character class, returning `null` with a `console.warn`, consistent with the converter's existing explicit rejection paths. Out-of-order ranges like `a-Z` (a `SyntaxError` in a real regexp engine) are rejected as a side effect.
- Keep the existing literal-character treatment for ranges with non-alphanumeric endpoints such as `!-?`.
- Protect `[a-zxA-Z-y]` as a must-still-convert case so validation does not invent a phantom `x-y` range after stripping supported ranges.
- Add regression coverage for both rejection paths and the must-still-convert path.

Rejecting the pattern avoids silently broadening its meaning or introducing a literal-hyphen requirement.

## Verification

Ran `sh tools/PatternRegexpToPasswordRulesConverter/run-tests.sh`; all converter tests pass, including both partial-range regressions and the full-range/literal-hyphen guard.

## Actual screenshots

| Before (`91312a9`) | After (`b5d48ad`) |
| --- | --- |
| Before: terminal reproduction showing partial ranges converted into literal-hyphen rules | After: terminal reproduction showing partial ranges rejected, the guard preserved, and all 52 tests passing |

```text
Test: Full ranges beside a literal hyphen
Expected: allowed: lower, upper, [-]; minlength: 8;
Got:      allowed: lower, upper, [-]; minlength: 8;
Status:   ✅ PASS

Test: Partial range in lookahead
Result:   null (rejected)
Status:   ✅ PASS

Test: Partial range in main character class
Result:   null (rejected)
Status:   ✅ PASS

All 52 tests passed.
```

### DIFF
--- a/tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverter.js
+++ b/tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverter.js
@@ -104,6 +104,12 @@ class PatternRegexpToPasswordRulesConverter {
             return null;
         }
 
+        if (this.#containsUnsupportedAlphanumericRange(charClassMatch[1])) {
+            console.warn("PatternRegexpToPasswordRulesConverter: Main character class contains an unsupported alphanumeric range");
+            console.warn("Pattern:", regexp);
+            return null;
+        }
+
         // Parse the validated components
         const required = this.#parseLookaheads(lookaheads);
         const allowed = this.#parseCharacterClassFromMatch(charClassMatch, required);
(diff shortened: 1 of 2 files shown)


rmondello (approved):

rmondello:
Thanks @Kevinjohn! Nice work.

PR #1168, opened 2026-07-20
Add a password rule for qualtrics.com

rmondello:
If you type fewer than eight characters, you get an error. If you don't include the right special character, you get:
Password must contain one or more special characters (e.g. ! @ # $ %)
    
Lower, upper, and digit are all allowed via testing.

### Overall Checklist

#### for password-rules.json

### DIFF
--- a/quirks/password-rules.json
+++ b/quirks/password-rules.json
@@ -983,6 +983,9 @@
     "qdosstatusreview.com": {
         "password-rules": "minlength: 8; required: lower; required: upper; required: digit; required: [!#$%&@^];"
     },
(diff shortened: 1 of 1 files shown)


aabosh:
It looks like this PR also includes the changes from https://github.com/apple/password-manager-resources/pull/1145

I think you meant to separate those out?

rmondello:
> It looks like this PR also includes the changes from #1145
> 
> I think you meant to separate those out?

Oops! Fixed.

aabosh (approved):

PR #1213, opened 2026-08-08
Reject character classes the converter cannot represent

### LINKED ISSUE #1212: Unsupported character classes are converted instead of rejected
Character classes the converter cannot represent are silently converted instead of rejected, producing rules that do not match the original pattern.

## Reproduction

Verified against `08f68a0`.

Negated class, required-character path:

- Pattern: `^(?=.*[0-9])(?=.*[^a-zA-Z0-9])[a-zA-Z0-9!@#$]{8,64}$`
- Expected: `null`, because a negated class is not representable.
- Actual: `required: digit; required: digit, upper, lower, [^]; allowed: [!@#$]; minlength: 8; maxlength: 64;`

The `^` is read as a member of the class rather than as negation, so "must contain a character that is **not** alphanumeric" becomes "must contain a digit, uppercase, lowercase, or a literal `^`". A generated `Abcdefg1` satisfies the rule and is rejected by the site. A smaller case shows the same inversion:

- Pattern: `^(?=.*[^0-9])[a-zA-Z0-9]{8,}$`
- Expected: `null`
- Actual: `required: digit, [^]; allowed: lower, upper; minlength: 8;`

Escape with an alphanumeric payload, allowed-character path:

- Pattern: `^[a-zA-Z0-9\s]{8,20}$`
- Expected: `null`, matching how the same escape is already handled in a lookahead.
- Actual: `allowed: lower, upper, digit, [\]; minlength: 8; maxlength: 20;`

`\s` collapses to a literal backslash, so the rule permits a character the site does not. This one is a path inconsistency rather than a new defect class: the lookahead path already rejects it, and returns `null` with `Lookaheads contain unsupported patterns` for `^(?=.*[a-z\s])[a-zA-Z0-9]{8,20}$`. The main character class accepts it silently. Normalization strips `[a-zA-Z0-9]` before escapes are unescaped, so any escape whose payload is alphanumeric (`\s`, `\S`, `\W`, `\t`, `\xNN`) leaves the backslash behind.

Enumerated class in a lookahead:

- Pattern: `^(?=.*[0123456789])[a-zA-Z0-9]{8,20}$`
- Expected: `required: digit; allowed: lower, upper, digit; minlength: 8; maxlength: 20;`, or `null`.
- Actual: `allowed: lower, upper, digit; minlength: 8; maxlength: 20;`, with `We shouldn't get here, in #parseLookaheads.` logged to the console.

The requirement is dropped entirely and the rule is emitted anyway, so a password with no digit satisfies it.

## Note

In each case the converter produces a well-formed rule that parses cleanly, so nothing downstream flags it; the output simply does not mean what the pattern meant. Returning `null` for these, as #1161 established for partial ranges, seems more consistent than converting them.

Happy to open a pull request if this is worth fixing. `[^A-Za-z0-9]` maps closely onto the existing `special` identifier, so the negated-class case could plausibly be converted rather than rejected, but that is a larger decision than the rest and I did not want to assume it.

erwinzhang7:
Character classes the converter cannot represent are silently converted instead of rejected, producing rules that do not match the original pattern.

### Overall Checklist

- [x] The top-level JSON objects are sorted alphabetically (N/A — no JSON changes)

Fixes #1212. Independent of #1160 and #1162, though it follows the rejection approach both established.

## Reproduction

Verified against `08f68a0`.

Negated class, both paths:

- Pattern: `^(?=.*[0-9])(?=.*[^a-zA-Z0-9])[a-zA-Z0-9!@#$]{8,64}$`
- Expected: `null`, because negation is not representable.
- Actual: `required: digit; required: digit, upper, lower, [^]; allowed: [!@#$]; minlength: 8; maxlength: 64;`

- Pattern: `^[^a-z]{8,}$`
- Expected: `null`
- Actual: `allowed: lower, [^]; minlength: 8;` — the inverse of the pattern.

Escape with an alphanumeric payload, main character class:

- Pattern: `^[a-zA-Z0-9\s]{8,20}$`
- Expected: `null`, matching the lookahead path, which already rejects `\s`.
- Actual: `allowed: lower, upper, digit, [\]; minlength: 8; maxlength: 20;`

Enumerated alphanumeric class in a lookahead:

- Pattern: `^(?=.*[0123456789])[a-zA-Z0-9]{8,20}$`
- Expected: `required: digit; ...` or `null`.
- Actual: `allowed: lower, upper, digit; minlength: 8; maxlength: 20;`, with `We shouldn't get here, in #parseLookaheads.` logged and the requirement dropped.

## Change

- Add `#containsUnrepresentableClass`, checked for both lookaheads and the main character class, returning `null` with a `console.warn` consistent with the converter's existing rejection paths.
- Treat a **leading** `^` as negation and reject it. A `^` elsewhere in a class stays a supported literal, which matters because shipped rules in `password-rules.json` use it.
- Reject escapes whose payload is alphanumeric, since normalization strips `[a-zA-Z0-9]` before escapes are unescaped and leaves a bare backslash. `\d` and `\w` are expanded by the converter and remain supported, including in bracket form.
- Add `#lookaheadHasNoRepresentableRequirement` for enumerated alphanumeric classes, which resolve to neither a standard range nor a non-alphanumeric residue. This replaces the `We shouldn't get here` path, where the requirement was dropped but the rule still emitted.
- Add regression coverage for all four rejections, plus two must-still-convert guards: literal mid-class `^` and `\d` inside the main character class.

## Verification

Ran `sh tools/PatternRegexpToPasswordRulesConverter/run-tests.sh`: 59 tests pass, from 53 before.

Running the six new cases against the unpatched converter fails 4 of 59 — the four rejection cases — so the new tests exercise the change rather than merely accompanying it. The two must-still-convert cases pass both before and after, which is what makes them useful as guards.

## Note

`[^A-Za-z0-9]` maps closely onto the existing `special` identifier, which `PasswordRulesParser.js` supports and `password-rules.json` uses, so that particular negation could plausibly be converted rather than rejected. I have not done that here: it is a larger decision about scope than the rest of this change, and rejecting is the conservative option. Happy to follow up if you would prefer it converted.

### DIFF
--- a/tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
+++ b/tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
@@ -497,6 +497,36 @@ class PatternRegexpToPasswordRulesConverterTest {
                 name: "Partial range in main character class",
                 regexp: "^[a-z2-9]{8,}$",
                 shouldBeNull: true
+            },
+            {
+                name: "Negated class in lookahead",
+                regexp: "^(?=.*[0-9])(?=.*[^a-zA-Z0-9])[a-zA-Z0-9!@#$]{8,64}$",
+                shouldBeNull: true
+            },
+            {
+                name: "Negated class in main character class",
+                regexp: "^[^a-z]{8,}$",
+                shouldBeNull: true
+            },
+            {
+                name: "Literal caret is not negation",
+                regexp: "^[a-zA-Z0-9!@#$%^&*]{8,}$",
+                shouldBeNull: false
+            },
+            {
+                name: "Whitespace escape in main character class",
+                regexp: "^[a-zA-Z0-9\\s]{8,20}$",
+                shouldBeNull: true
+            },
+            {
+                name: "Digit shorthand in main character class",
+                regexp: "^[a-z\\d]{8,}$",
+                shouldBeNull: false
+            },
+            {
+                name: "Enumerated alphanumeric class in lookahead",
+                regexp: "^(?=.*[0123456789])[a-zA-Z0-9]{8,20}$",
+                shouldBeNull: true
             }
         ];
 
--- a/tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
+++ b/tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
@@ -295,6 +295,11 @@ class PatternRegexpToPasswordRulesConverterTest {
                 name: "Both hyphen and bracket (escaped) in lookahead",
                 regexp: "^(?=.*[\\]-!@])(?=.*[0-9])[a-zA-Z0-9\\]-!@]{8,20}$",
                 expected: "required: [-!@]]; required: digit; allowed: lower, upper; minlength: 8; maxlength: 20;"
+            },
+            {
+                name: "Caret is a literal when it does not lead the character class",
+                regexp: "^[a-zA-Z0-9!@#$%^&*]{8,}$",
+                expected: "allowed: lower, upper, digit, [!@#$%^&*]; minlength: 8;"
             }
         ];
 
@@ -434,6 +439,11 @@ class PatternRegexpToPasswordRulesConverterTest {
                 name: "Escaped backslash in main character class",
                 regexp: "^(?=.*[0-9])[a-z0-9\\\\]{8,20}$",
                 expected: "required: digit; allowed: lower, [\\]; minlength: 8; maxlength: 20;"
+            },
+            {
+                name: "\\d in main character class without a lookahead",
+                regexp: "^[a-z\\d]{8,}$",
+                expected: "allowed: lower, digit; minlength: 8;"
             }
         ];
 
@@ -509,19 +519,9 @@ class PatternRegexpToPasswordRulesConverterTest {
                 shouldBeNull: true
             },
             {
-                name: "Literal caret is not negation",
-                regexp: "^[a-zA-Z0-9!@#$%^&*]{8,}$",
-                shouldBeNull: false
-            },
-            {
                 name: "Whitespace escape in main character class",
                 regexp: "^[a-zA-Z0-9\\s]{8,20}$",
                 shouldBeNull: true
-            },
-            {
-                name: "Digit shorthand in main character class",
-                regexp: "^[a-z\\d]{8,}$",
-                shouldBeNull: false
             },
             {
                 name: "Enumerated alphanumeric class in lookahead",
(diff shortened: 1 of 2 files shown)


tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
```diff
@@ -497,6 +497,36 @@ class PatternRegexpToPasswordRulesConverterTest {
...
+            {
+                name: "Negated class in main character class",
+                regexp: "^[^a-z]{8,}$",
+                shouldBeNull: true
+            },
+            {
+                name: "Literal caret is not negation",
+                regexp: "^[a-zA-Z0-9!@#$%^&*]{8,}$",
+                shouldBeNull: false
+            },
```
rmondello: The rest of the tests in this section are `shouldBeNull: true`. This leads me to believe this test case would be better off somewhere else in the file.

tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
```diff
@@ -497,6 +497,36 @@ class PatternRegexpToPasswordRulesConverterTest {
...
+            {
+                name: "Whitespace escape in main character class",
+                regexp: "^[a-zA-Z0-9\\s]{8,20}$",
+                shouldBeNull: true
+            },
+            {
+                name: "Digit shorthand in main character class",
+                regexp: "^[a-z\\d]{8,}$",
+                shouldBeNull: false
+            },
```
rmondello: The rest of the tests in this section are `shouldBeNull: true`. This leads me to believe this test case would be better off somewhere else in the file.

rmondello (changes requested):

rmondello:
Other than where those test cases live (or a discussion about that being the right place), change looks good to me.

tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
erwinzhang7: Moved to testCharacterClassNormalization, next to the other tests about how characters inside a class are parsed. It now asserts the exact output rather than just non-null.

tools/PatternRegexpToPasswordRulesConverter/PatternRegexpToPasswordRulesConverterTest.js
erwinzhang7: Moved to testShorthandAndCombinedPatterns, beside the existing \d case. Also asserts the exact output now.

rmondello (approved):

rmondello:
@erwinzhang7 Just curious: what brought you to hacking on this? Are you using this code in a project somewhere, or more just looking for something in open source to hack on?

rmondello:
@erwinzhang7 Thank you for the contribution!

erwinzhang7:
Using it, I pulled the rules data into an accounting app for my own bookkeeping, and ran the converter over the corpus while I was in there. A few outputs didn't mean what the source pattern meant, and the negated-class one was the worst of them.

Thanks for the quick review.

